PR #1, opened 2020-06-29
Add coverage and pytest to CI

Hofer-Julian:
Also adds a dummy test and a dummy example

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -20,11 +20,21 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install imod_coupler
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -20,10 +20,19 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install imod_coupler
+      - name: Install dependencies
```
rleander, line 23: Why do you install dependencies?

Hofer-Julian: Because the dependencies are necessary for testing

rleander:
Including Julian's commits

PR #4, opened 2020-07-01
Revert "Fixing the read map function for the specific mod2svat.inp fo…

JoerivanEngelen:
This reverts commit 0097d12c7c48ac5b7acd4f8901ba151f3a2f8122.

I discussed this with Robert, and the strip model example has a mod2svat.inp with a format that does not really conform to what is "normal" for MetaSWAP. All files I've seen, except the mod2svat.inp in the strip model, pad with spaces and do not use comma'
s and quotes as seperators. 

I think this specfic mod2svat.inp file in the strip model creeped in during the development of the coupler script, as it is easier to parse with regex, which was used initially to read input files in this python script. 

It is therefore probably better to fix the mod2svat.inp file in the strip model example instead and conform to what is usual than to force others to stick to our specific format.

tl;dr: It is best to revert this commit ;-)

### DIFF
--- a/imod_coupler/metamod.py
+++ b/imod_coupler/metamod.py
@@ -26,12 +26,17 @@ def mapids(ids1, ids2):
 
 
 def read_mapping(map_file):
-    map_arr = np.loadtxt(map_file, dtype=str, delimiter=",")
+
+    map_arr = np.loadtxt(map_file, dtype=np.int32)
+    map_arr[:, 0] = (
+        map_arr[:, 0] - 1
+    )  # 1-based indices (fortran) to 0-based indices (python)
+    map_arr[:, 1] = map_arr[:, 1] - 1
+
     map_in = {}
-    for sgw, ssv in map_arr:
-        gw = int(sgw) - 1
-        sv = int(ssv.replace("'", "").split()[0])
-        map_in.setdefault(gw, []).append(sv - 1)
+    for gw, sv in zip(map_arr[:, 0], map_arr[:, 1]):
+        map_in.setdefault(gw, []).append(sv)
+
     return map_in
 
 
@@ -226,3 +231,4 @@ class MetaMod(AmiWrapper):
         # because the condition "if i in map_msw2mod" is never true.
         # Initalize the heads in MetaSWAP by copying them from MODFLOW
         self.xchg_mod2msw()
+


rleander:
> This reverts commit 0097d12.
> I discussed this with Robert, and the strip model example has a mod2svat.inp with a format that does not really conform to what is "normal" for MetaSWAP. All files I've seen, except the mod2svat.inp in the strip model, pad with spaces and do not use comma'
> s and quotes as seperators.
> I think this specfic mod2svat.inp file in the strip model creeped in during the development of the coupler script, as it is easier to parse with regex, which was used initially to read input files in this python script.
> It is therefore probably better to fix the mod2svat.inp file in the strip model example instead and conform to what is usual than to force others to stick to our specific format.
> tl;dr: It is best to revert this commit ;-)

The assumption that this is a newly produced format (i.e. comes with the setup of the script) seems to be disproven by inspection of Paul's original strip model, of which the input files are stored at

https://repos.deltares.nl/repos/DSCTestbench/trunk/cases/e150_metaswap/f01_basic_tests/c01_ModStrip1/testunsa/inp ... also the code in SIMGRO seems to be prepared  `READ(iun,'(i10,2x,i10,i5)',IOSTAT=ios) mf_r,k_r,ly_r`:

JoerivanEngelen:
Ah interesting! So that hypothesis goes out of the window.

I'm not proficient in FORTRAN, but it seems to me that MetaSWAP does not care about what characters are used as seperators. So whether it is:
```
         2,'         1    1'
         3,'         2    1'
```

or:

```
         2           1    1 
         3           2    1 
```

In my test model, MetaSWAP handled the latter format fine.
However, the iMOD-converter writes files only in the latter format, and for reading and writing files in Python, it is also a lot easier to use the latter format. So I am still inclined towards supporting the latter format in our coupler script. I also cannot find in the IO manual why you would use these comma's and quotes as seperators.

rleander:
The 2X in the read format absorbs the 
,'

or anything. So effectively with the 2X fortran reads the same line as if it were without the punctuation marks. That's why it works. I didn't even consider it yesterday, just modified the script to make it work, because I was in a rush. But I agree with you it is better to leave it this way. Furthermore, the last column lost its function I think... we are not using it.

PR #5, opened 2020-07-01
Add AGPL licence

Hofer-Julian:


mjr-deltares:
Hi, just wanted to let you know that this is in progress.

Hofer-Julian:
We will use LGPL instead.

PR #21, opened 2020-08-07
Sparse exchange

rleander:
I have rhe new exchange routine working using sparse matrices, tested it quantitatively by comparing with the arrays after exchange in the old and new method and ran flak8. Please review my branch


rleander:
I still have to add all the descriptions and doxygen annotations, but I will do that in a separate branch

mjr-deltares:
and as we just learned that there is a huge speedup of the time spent in the coupler, there is no reason to hold this back ;-)

PR #22, opened 2020-08-11
Config toml

rleander:
Can anyone please rewview this? I already tested with a working json config for the f01_c20_Hooghoudt test model

{
    "components": {
        "mf6": {
            "dll": "d:\\leander\\My_Projects\\ModSWAPTransMoz\\collector\\libmf6.dll",
            "engine": "mf6",
            "wd": "."
            },
        "msw": {
             "dll": "d:\\leander\\My_Projects\\ModSWAPTransMoz\\collector\\MetaSWAP.dll",
             "engine": "msw",
             "wd": "GWF_1\\MSWAPINPUT"
             }
        },
    "dependencies": ["d:\\leander\\My_Projects\\ModSWAPTransMoz\\collector"],
    "exchanges": [],
    "parameters": []
}

### DIFF
--- a/imod_coupler/__main__.py
+++ b/imod_coupler/__main__.py
@@ -1,5 +1,6 @@
 import argparse
 import logging
+import json
 import os
 import sys
 import time
@@ -83,6 +84,13 @@ def main():
     )
 
     parser.add_argument(
+        "--config",
+        action="store",
+        default="",
+        help="Provide optional config file in json-format.",
+    )
+
+    parser.add_argument(
         "--timing",
         action="store_true",
         help="activates timing, verbosity can be adjusted with the log-level.",
@@ -101,9 +109,12 @@ def main():
     msw_dll = args.msw_dll
     mf6_model_dir = args.mf6_model_dir
     msw_model_dir = args.msw_model_dir
+    configfile = args.config
     msw_mpi_dll_dir = args.msw_mpi_dll_dir
     debug_native = args.enable_debug_native
     timing = args.timing
+    print('Config : '+configfile)
+    sys.exit()
 
     if timing:
         start = time.perf_counter()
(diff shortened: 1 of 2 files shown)


imod_coupler/__main__.py
```diff
@@ -29,31 +30,31 @@ def main():
 
     parser.add_argument(
```
Hofer-Julian, line 31: I think, we want to get rid of command line options and env variables completely as soon as we have a config file

imod_coupler/__main__.py
```diff
@@ -94,53 +103,53 @@ def main():
         version="%(prog)s {version}".format(version=__version__),
     )
 
+    if len(sys.argv) == 1:
+        sys.argv.append("--help")
```
Hofer-Julian, line 107: This should not be necessary.
Have you tested this?

imod_coupler/__main__.py
```diff
@@ -94,53 +103,53 @@ def main():
...
-    if not os.path.isdir(msw_mpi_dll_dir):
-        logger.error(f"Metaswap MPI dlls {msw_mpi_dll_dir} not found.")
-        sys.exit(1)
-
     # wait for native debugging
     if debug_native:
         input(f"PID: {os.getpid()}, press any key to continue ....")
 
+    config_data = \
+        {'components': {}, 'dependencies': [], 'parameters': [], 'exchanges': []}
```
Hofer-Julian, line 128: I think we want to have a Config class, instead of a dict here.
For inspiration you can see https://github.com/anoadragon453/nio-template/blob/master/my_project_name/config.py

Hofer-Julian: I also think that the config logic should not be done in __main__ but rather in metamod, so that it can be used as a library. So e.g. metamod should receive a path to the config file and then does the rest itself.

Hofer-Julian:
https://github.com/rochacbruno/dynaconf seems to be suitable

PR #24, opened 2020-08-26
Crucial fixes to the metamod.py script:

mjr-deltares:
- SC1 should be used, not SC2
- there is a reset flag that should be set whenever the value changes to trigger
modflow side effects
- Recharge is the first column in the BOUND array, which has rank 2!
- there's no head exchange into metaswap inside the initialize, BUT,
there should be one before the iteration loop
- array assignment results causes the member (e.g. self.mf6_head) to no
longer point to the proper array inside the Fortran kernel, replaced this
by broadcast

This addresses issue IMOD6-271, which should be tested again.

(Paired with RL)

### DIFF
--- a/imod_coupler/metamod.py
+++ b/imod_coupler/metamod.py
@@ -136,7 +136,7 @@ class MetaMod:
         mf6_max_iter_tag = self.mf6.get_var_address("MXITER", "SLN_1")
 
         self.mf6_head = self.mf6.get_value_ptr(mf6_head_tag)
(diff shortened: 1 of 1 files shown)


rleander (approved):
Review OK

Hofer-Julian:
The testbench fails now.
This is expected I guess since it compares to the old (and wrong?) values.
Let's discuss this tomorrow :)

rleander:
> The testbench fails now.
> This is expected I guess since it compares to the old (and wrong?) values.
> Let's discuss this tomorrow :)

That is expected as results are different and actually make sense (I hope). Good to see that out testbench detects it.
Since we know know that our old results do NOT make sense, I can replace references with the new results

PR #25, opened 2020-08-27
Refactor imod_coupler

Hofer-Julian:
- Move initialize as discussed with Robert and Martijn
- Make timing and log_level config option
- Remove Kernel and replace with XMIWrapper

Please see https://issuetracker.deltares.nl/browse/IMOD6-357


rleander (approved):
Review OK (Kernel class was not needed, now directly using XMIWrapper instances)

PR #28, opened 2020-08-31
Finalize CLI interface

Hofer-Julian:
After a talk with Joeri, I try to to finalize the command line interface with this PR.
New options can be added, but existing ones should not be modified anymore.

This PR includes
- timing and log_level is moved to CLI again
- config_path is a positional parameter now
- refactor timing (otherwise it would stop working
   when we add additional kernels)

Instead of
```
imodc --config-path=/path/to/imod_coupler.toml
```
it is now
```
imodc /path/to/imod_coupler.toml
```

The necessary changes are already done in the testbench, so as soon as this is merged it should work again.

The current design is:
- user- (or developer-) specific settings via the CLI
- model settings via the config file (which can be overriden via env variables)

If you have any last design suggestions, please free to add them :)

JIRA issue can be found here https://issuetracker.deltares.nl/browse/IMOD6-362

### DIFF
--- a/imod_coupler/__main__.py
+++ b/imod_coupler/__main__.py
@@ -16,9 +16,29 @@ def main():
     parser = argparse.ArgumentParser()
 
     parser.add_argument(
+        "config_path",
+        action="store",
+        help="specify the path to the configuration file",
+    )
+
+    parser.add_argument(
+        "--log-level",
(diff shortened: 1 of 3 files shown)


rleander (approved):
Review OK

PR #29, opened 2020-09-02
Global environment variables in toml-section at Joeri's request

rleander:
Joeri requested the possibility to add global environment variables to the toml file. I added and tested it. Please discuss with Joeri and decided if that is still desired. To be honest, I was against it: users should open a dosbox to set the environment

### DIFF
--- a/imod_coupler/config.py
+++ b/imod_coupler/config.py
@@ -33,6 +33,10 @@ class Config(object):
 
             self.get_kernel_data(kernels, timing)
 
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Yes, let's discuss this tomorrow.
At least after a quick thought, I don't really see the use case here.

rleander:
OK, neither do I.. then lets forget about it, we have better things to do

Hofer-Julian:
Closing, as discussed in today's meeting

PR #30, opened 2020-09-07
Use pathlib in cd contextmanager

Hofer-Julian:
Otherwise relative paths would only be accepted when they start with "./"

Please see https://issuetracker.deltares.nl/browse/IMOD6-366


mjr-deltares (approved):

mjr-deltares:
Testbench Coupler Win64 is failing, but this seems to be unrelated. Will merge.

PR #31, opened 2020-09-07
Fix flow direction

JoerivanEngelen:
Changed the sign of the `qmodf` flow between MF6 and MetaSWAP. 

I implemented this incorrectly before, to account for errors that were apparently created by the diverging heads. 

The head divergence issue is now fixed in the MetaSWAP.dll and it seems that multiplying `qmodf` with -1.0 created an erronous coupling and should therefore be reverted.


mjr-deltares (approved):

mjr-deltares:
let's fix black and then merge...

Hofer-Julian:
Formatted with black and merged

PR #32, opened 2020-09-08
metaswap properly converts the storage coefficient

mjr-deltares:
before exchanging with modflow, no need to trigger the update routine inside modflow.

https://issuetracker.deltares.nl/browse/IMOD6-384


JoerivanEngelen (approved):
Seems alright to me.

PR #37, opened 2020-10-06
Imod6 361 sprinkling

mjr-deltares:


JoerivanEngelen (approved):
Looks good, I have not it tested yet, but it looks like it should work. And otherwise I'll raise a next issue ;-)

PR #39, opened 2020-10-12
- made sprinkling optional

mjr-deltares:
- add proper tag for MSW-RCH variable


JoerivanEngelen:
Code looks good, apparently required for me an update of XMIPY and the .dll locally.

JoerivanEngelen (approved):
Code looks good, apparently required for me an update of XMIPY and the .dll locally.

PR #43, opened 2021-05-06
Update guide-to-publish.md

Hofer-Julian:
I've found that without removing .egg-info files things can go wrong


mjr-deltares:
a bit late... but review ok ;-)

PR #65, opened 2022-04-29
Imod6-848: Setup TeamCity

Hofer-Julian:


imod_coupler/drivers/driver.py
```diff
@@ -0,0 +1,54 @@
+""" MetaMod: the coupling between MetaSWAP and MODFLOW 6
```
mjr-deltares, line 1: PS: this comment appears to be a relic

Hofer-Julian: Indeed, you found some copypasta there.

imod_coupler/drivers/metamod/metamod.py
```diff
@@ -14,12 +14,13 @@
...
 from xmipy import XmiWrapper
 
+from imod_coupler.drivers.driver import Driver
 from imod_coupler.utils import create_mapping
 
 logger = logging.getLogger(__name__)
 
 
-class MetaMod:
+class MetaMod(Driver):
```
JoerivanEngelen, line 23: I think it would be good to add some docstrings here of what the "MetaMod" object takes care off:
It provides a driver to couple MetaSWAP to Modflow 6.

JoerivanEngelen (changes requested):
My only comment is a bit outside the scope of the intention of this merge request, but since the driver was also generalized in this specific pull request, I still took the liberty to request a few extra docstrings ;-).

imod_coupler/drivers/metamod/metamod.py
Hofer-Julian: Fixed

PR #69, opened 2022-05-04
Imod6-849

JoerivanEngelen:
For issue, see https://issuetracker.deltares.nl/browse/IMOD6-849

Requires this branch from imod-python:
https://gitlab.com/deltares/imod/imod-python/-/merge_requests/132

### DIFF
--- a/tests/test_metamod.py
+++ b/tests/test_metamod.py
@@ -258,4 +258,6 @@ def test_metamod_regression_sprinkling(
     assert budgets_dev.keys() == budgets_reg.keys()
 
     for varname in budgets_dev.keys():
-        assert_array_almost_equal(budgets_dev[varname].compute(), budgets_reg[varname].compute(), decimal=8)
+        assert_array_almost_equal(
+            budgets_dev[varname].compute(), budgets_reg[varname].compute(), decimal=8
+        )


tests/test_metamod.py
```diff
@@ -25,3 +31,37 @@ def test_metaswap_dll_dep_dir_contains_dependencies(metaswap_dll_dep_dir: Path)
...
+    )
+
+    metamod.write(
+        tmp_path,
+        modflow6_dll=modflow_dll,
+        metaswap_dll=metaswap_dll,
+        metaswap_dll_dependency=metaswap_dll_dep_dir,
+    )
+
+    os.chdir(tmp_path)
```
Hofer-Julian, line 64: Chdir is not necessary, you can just specify the full path of the toml

tests/test_metamod.py
```diff
@@ -25,3 +31,37 @@ def test_metaswap_dll_dep_dir_contains_dependencies(metaswap_dll_dep_dir: Path)
...
+    subprocess.run(["imodc", "metamod.toml"])
```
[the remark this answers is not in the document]
Hofer-Julian: Yes, I think naming the config file consistently `imod_coupler.toml` makes it easier to identify it.

tests/test_metamod.py
```diff
@@ -1,6 +1,70 @@
...
+
+
+@pytest.fixture
+def tmp_path_reg(
+    tmp_path: Path,
+) -> Path:
+    return tmp_path / "regression"
+
+
+def read_log_for_succes(logfile_path):
```
Hofer-Julian, line 46:
```suggestion
def read_log_for_success(logfile_path):
```

tests/test_metamod.py
```diff
@@ -1,6 +1,70 @@
...
+                has_msw_succes_message = True
+
+    return has_mf6_succes_message, has_msw_succes_message
+
+
+def run_model(path: Path, imod_coupler_exec: Path):
+    logfile_path = path / "metamod.log"
+    with open(logfile_path, "w") as logfile:
+        subprocess.run(
+            [str(imod_coupler_exec), path / "metamod.toml"],
```
Hofer-Julian, line 63:
```suggestion
            [imod_coupler_exec, path / "imod_coupler.toml"],
```

tests/test_metamod.py
```diff
@@ -1,6 +1,70 @@
...
+            stdout=logfile,
```
Hofer-Julian, line 64: Why not just parse MODFLOW/MetaSWAP logs?

tests/test_metamod.py
```diff
@@ -27,3 +91,120 @@ def test_metaswap_dll_dep_dir_contains_dependencies(
...
+    )
+
+    # Capture standard output and error in file (instead of StringIO) for
+    # debugging purposes
+    run_model(tmp_path_reg, imod_coupler_exec_regression)
+
+    has_mf6_succes_message, has_msw_succes_message = read_log_for_succes(
+        tmp_path_reg / "metamod.log"
+    )
+    assert has_mf6_succes_message
```
Hofer-Julian, line 194:
```suggestion
    assert has_mf6_success_message
```

tests/test_metamod.py
```diff
@@ -27,3 +91,120 @@ def test_metaswap_dll_dep_dir_contains_dependencies(
...
+    assert has_msw_succes_message
```
Hofer-Julian, line 195:
```suggestion
    assert has_msw_success_message
```

Hofer-Julian (commented):
A few small things I've noticed

tests/test_metamod.py
```diff
@@ -1,6 +1,70 @@
...
+
+
+def read_log_for_succes(logfile_path):
+    with open(logfile_path, "r") as logfile:
+        has_mf6_succes_message = False
+        has_msw_succes_message = False
+        for line in logfile:
+            if "Normal termination of simulation." in line:
+                has_mf6_succes_message = True
+            if "E N D   O F   M E T A S W A P" in line:
```
Hofer-Julian, line 53: Parsing the log like that shouldn't be necessary.
I would expect that a failure in run would result in an exist code of 1 of imodc, which would throw an exception in subprocess and therefore fail the test. If you see a different behavior I would consider that a bug

JoerivanEngelen: We discussed this, and we thought it was still to keep it in the test for now, as we are yet unsure that ```imodc``` returns a exit code of 1 upon model non-convergence or other failures, because the DLLs do not return exit codes.

tests/test_metamod.py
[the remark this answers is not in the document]
JoerivanEngelen: Done

tests/test_metamod.py
JoerivanEngelen: Removed the chdir

tests/test_metamod.py
JoerivanEngelen: Fixed typo

tests/test_metamod.py
JoerivanEngelen: As this is the first thing a user is confronted with upon failure/success, apart from that not much.

tests/test_metamod.py
JoerivanEngelen: Fixed typo

tests/test_metamod.py
JoerivanEngelen: Changed the name to ``imod_coupler.toml``

tests/test_metamod.py
```diff
@@ -1,6 +1,70 @@
...
+
+
+@pytest.fixture
+def metamod_model_sprinkling(
+    coupled_mf6_model: Modflow6Simulation,
+    msw_model: MetaSwapModel,
+    metaswap_lookup_table: Path,
+) -> MetaMod:
+
+    # Override unsat_svat_path with path from .env
```
Hofer-Julian, line 19:
```suggestion
    # Override unsat_svat_path with path from environment
```

The .env just sets the environment

tests/test_metamod.py
```diff
@@ -1,6 +1,70 @@
...
+                has_msw_success_message = True
+
+    return has_mf6_success_message, has_msw_success_message
+
+
+def run_model(path: Path, imod_coupler_exec: Path):
+    logfile_path = path / "metamod.log"
+    with open(logfile_path, "w") as logfile:
+        subprocess.run(
+            [str(imod_coupler_exec), path / "imod_coupler.toml"],
```
Hofer-Julian, line 63:
```suggestion
            [imod_coupler_exec, path / "imod_coupler.toml"],
```

Hofer-Julian (commented):
Looks good to me apart from two minor things

JoerivanEngelen:
I added a pytest-cases dependency, as this package allows very nicely to seperate cases from tests, which is something that bugged me. This would make it a lot easier to add extra cases to our tests, without code duplication. See https://smarie.github.io/python-pytest-cases/

tests/test_metamod.py
```diff
@@ -50,3 +136,128 @@ def test_metaswap_dll_dep_dir_regression_contains_dependencies(
...
+    grbfile_reg = tmp_path_reg / "Modflow6" / "GWF_1" / "dis.dis.grb"
+
+    heads_reg = open_hds(headfile_reg, grbfile_reg)
+    budgets_reg = open_cbc(cbcfile_reg, grbfile_reg)
+
+    assert_array_almost_equal(heads_dev.compute(), heads_reg.compute(), decimal=8)
+
+    assert budgets_dev.keys() == budgets_reg.keys()
+
+    ## FUTURE: Activate when issue is fixed:
```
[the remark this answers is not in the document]
JoerivanEngelen: Good catch! I forgot about that one, I'll unccoment it and see if the tests still work

PR #72, opened 2022-08-15
Add storage coefficient test

JoerivanEngelen:
Added a testcase and test for storage coefficient.

### DIFF
--- a/tests/test_metamod.py
+++ b/tests/test_metamod.py
@@ -7,47 +7,40 @@ from imod.couplers.metamod import MetaMod
 from imod.mf6 import Modflow6Simulation, open_cbc, open_hds
 from imod.msw import MetaSwapModel
 from numpy.testing import assert_array_almost_equal
-from pytest_cases import parametrize_with_cases
+from pytest_cases import parametrize_with_cases, parametrize, fixture
 
 
-def case_metamod_model_sprinkling(
-    coupled_mf6_model: Modflow6Simulation,
+@fixture(scope="function")
+def prepared_msw_model(
     msw_model: MetaSwapModel,
     metaswap_lookup_table: Path,
-) -> MetaMod:
-
+) -> MetaSwapModel:
     # Override unsat_svat_path with path from environment
     msw_model.simulation_settings[
         "unsa_svat_path"
     ] = msw_model._render_unsaturated_database_path(metaswap_lookup_table)
 
-    return MetaMod(
-        msw_model,
-        coupled_mf6_model,
-        mf6_rch_pkgkey="rch_msw",
-        mf6_wel_pkgkey="wells_msw",
-    )
+    return msw_model
 
 
-def case_metamod_model_no_sprinkling(
+@parametrize("use_sprinkling", [True, False], idstyle="explicit")
+def case_metamod_sprinkling(
     coupled_mf6_model: Modflow6Simulation,
-    msw_model: MetaSwapModel,
-    metaswap_lookup_table: Path,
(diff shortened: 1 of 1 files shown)


tests/test_metamod.py
```diff
@@ -261,3 +297,78 @@ def test_metamod_regression_sprinkling(
...
+
+    # Read Modflow 6 output
+    headfile_sc = tmp_path_sc / "Modflow6" / "GWF_1" / "GWF_1.hds"
+    cbcfile_sc = tmp_path_sc / "Modflow6" / "GWF_1" / "GWF_1.cbc"
+    grbfile_sc = tmp_path_sc / "Modflow6" / "GWF_1" / "dis.dis.grb"
+
+    heads_sc = open_hds(headfile_sc, grbfile_sc)
+    budgets_sc = open_cbc(cbcfile_sc, grbfile_sc)
+
+    assert_array_almost_equal(heads_ss.compute(), heads_ss.compute(), decimal=8)
```
Hofer-Julian, line 367: You compared two times heads_ss

tests/test_metamod.py
```diff
@@ -4,50 +4,86 @@
...
+
+def case_metamod_storage_coefficient(
+    coupled_mf6_model: Modflow6Simulation,
+    prepared_msw_model: MetaSwapModel,
+) -> MetaMod:
+
+    gwf_model = coupled_mf6_model["GWF_1"]
+
+    # Specific storage package
+    # gwf_model["sto"] = mf6.SpecificStorage(1e-3, 0.1, True, 0)
```
Hofer-Julian, line 62: Should that be commented out?

Hofer-Julian (changes requested):

tests/test_metamod.py
JoerivanEngelen: :( Good catch

tests/test_metamod.py
```diff
@@ -1,55 +1,161 @@
...
+    metamod_ss = MetaMod(prepared_msw_model, coupled_mf6_model, **kwargs)
+
+    metamod_sc = MetaMod(
+        prepared_msw_model, coupled_mf6_model_storage_coefficient, **kwargs
+    )
+
+    return metamod_ss, metamod_sc
+
+
+def storage_comparison_sprinkling(
```
Hofer-Julian, line 134: This one also doesn't compare anything.

tests/test_metamod.py
```diff
@@ -65,6 +171,13 @@ def tmp_path_reg(
     return tmp_path / "regression"
 
 
+def mf6_output_files(path: Path) -> Tuple[Path]:
+    """return paths to"""
```
Hofer-Julian, line 175: Return paths to... what?

tests/test_metamod.py
```diff
@@ -123,7 +236,7 @@ def test_modflow_dll_present(modflow_dll_devel: Path) -> None:
     assert modflow_dll_devel.is_file()
 
 
-@parametrize_with_cases("metamod_model", cases=".")
+@parametrize_with_cases("metamod_model", cases=".", prefix="case_")
 def test_metamod_develop(
```
Hofer-Julian, line 240: The name is not very descriptive.

But maybe instead of making the name longer it would be good to add docstrings to the tests?

How about we add docstrings to any non-trivial test?
Maybe longer than 10 lines?

Hofer-Julian (changes requested):
A few nitpicks, but fine overall :)

PR #74, opened 2022-08-16
Add failure cases and specific test to test if we catch failures.

JoerivanEngelen:
I added three cases where either the modflow 6 or metaswap failures should fail.

I couldn't come up with a case where metaswap did not converge, as even:
``iter_outer =  2 : [Hgwmsw - Hgwmodf] (svat=      1)=-99.99565 m``

was good enough for metaswap,

### DIFF
--- a/tests/test_metamod.py
+++ b/tests/test_metamod.py
@@ -116,7 +116,7 @@ def failure_msw_input(
     prepared_msw_model: MetaSwapModel,
 ) -> MetaMod:
     """
-    Force an input error in MetaSWAP by providing a initial condition with the
+    Force an input error in MetaSWAP by providing an initial condition with the
     wrong type value
     """
 
@@ -140,7 +140,7 @@ def failure_mf6_input(
     prepared_msw_model: MetaSwapModel,
 ) -> MetaMod:
     """
-    Force an input error in Modflow 6 by providing a k value of 0.0
+    Creates a MetaMod object which will result in a MODFLOW 6 input error.
     """
 
     coupled_mf6_model["GWF_1"]["npf"]["k"] *= 0.0
@@ -158,7 +158,7 @@ def failure_mf6_convergence(
     prepared_msw_model: MetaSwapModel,
 ) -> MetaMod:
     """
-    Force a non-convergencent solution, by providing extreme differences in the
+    Creates a MetaMod object which will result in a non-convergencent MODFLOW 6 solution, by providing extreme differences in the
     k values.
     """
 
@@ -319,7 +319,7 @@ def test_metamod_failure(
     imod_coupler_exec_devel: Path,
 ):
     """
-    Test if coupled models run fail with the iMOD Coupler development version.
+    Test if coupled models run faild with the iMOD Coupler development version.
     """
     metamod_model.write(
         tmp_path_dev,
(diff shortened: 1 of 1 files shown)


tests/test_metamod.py
```diff
@@ -110,6 +111,70 @@ def case_storage_coefficient_no_sprinkling(
     )
 
 
+def failure_msw_input(
+    coupled_mf6_model: Modflow6Simulation,
+    prepared_msw_model: MetaSwapModel,
+) -> MetaMod:
+    """
+    Force an input error in MetaSWAP by providing a initial condition with the
```
Hofer-Julian, line 119:
```suggestion
    Force an input error in MetaSWAP by providing an initial condition with the
```

tests/test_metamod.py
```diff
@@ -110,6 +111,70 @@ def case_storage_coefficient_no_sprinkling(
...
+    wrong type value
```
Hofer-Julian, line 120: "value of the wrong type" maybe?

tests/test_metamod.py
```diff
@@ -110,6 +111,70 @@ def case_storage_coefficient_no_sprinkling(
...
+    """
+
+    from imod.msw.fixed_format import VariableMetaData
```
Hofer-Julian, line 123: Is there a reason why you wouldn't put the import at the top of the file?

tests/test_metamod.py
```diff
@@ -110,6 +111,70 @@ def case_storage_coefficient_no_sprinkling(
...
+        mf6_wel_pkgkey="wells_msw",
+    )
+
+
+def failure_mf6_input(
+    coupled_mf6_model: Modflow6Simulation,
+    prepared_msw_model: MetaSwapModel,
+) -> MetaMod:
+    """
+    Force an input error in Modflow 6 by providing a k value of 0.0
```
Hofer-Julian, line 143:
```suggestion
    Creates a MetaMod object which will result in a MODFLOW 6 input error.
```

tests/test_metamod.py
```diff
@@ -110,6 +111,70 @@ def case_storage_coefficient_no_sprinkling(
...
+        mf6_wel_pkgkey="wells_msw",
+    )
+
+
+def failure_mf6_convergence(
+    coupled_mf6_model: Modflow6Simulation,
+    prepared_msw_model: MetaSwapModel,
+) -> MetaMod:
+    """
+    Force a non-convergencent solution, by providing extreme differences in the
```
Hofer-Julian, line 161:
```suggestion
    Creates a MetaMod object which will result in a non-convergencent MODFLOW 6 solution, by providing extreme differences in the
```

tests/test_metamod.py
```diff
@@ -244,6 +309,32 @@ def test_modflow_dll_present(modflow_dll_devel: Path) -> None:
...
+def test_metamod_failure(
+    tmp_path_dev: Path,
+    metamod_model: MetaMod,
+    metaswap_dll_devel: Path,
+    metaswap_dll_dep_dir_devel: Path,
+    modflow_dll_devel: Path,
+    imod_coupler_exec_devel: Path,
+):
+    """
+    Test if coupled models run fail with the iMOD Coupler development version.
```
Hofer-Julian, line 322:
```suggestion
    Test if coupled models run faild with the iMOD Coupler development version.
```

Hofer-Julian (changes requested):
Again only nitpicks.

I like these tests a lot!

PR #84, opened 2022-08-23
Specify relative dir at object level

Hofer-Julian:


imod_coupler/__main__.py
```diff
@@ -34,17 +34,16 @@ def run_coupler(config_path: Path) -> None:
     with open(config_path, "rb") as f:
         config_dict = tomllib.load(f)
 
-    # Validation expects current working directory at config file level
-    os.chdir(config_path.parent)
-    base_config = BaseConfig(**config_dict)
+    config_dir = config_path.parent
+    base_config = BaseConfig(config_dir=config_dir, **config_dict)
```
JoerivanEngelen, line 38: Could you repeat the comment here that the BaseConfig changes dir to ``config_dir`` at initialization here as well?

imod_coupler/__main__.py
```diff
@@ -34,17 +34,16 @@ def run_coupler(config_path: Path) -> None:
...
 
     logging.basicConfig(level=base_config.log_level)
 
     if base_config.timing:
         start = time.perf_counter()
 
     if base_config.driver_type == "metamod":
-        metamod_config = MetaModConfig(**config_dict["driver"])
+        metamod_config = MetaModConfig(config_dir=config_dir, **config_dict["driver"])
```
JoerivanEngelen, line 46: And here same as previous comment, but for MetaModConfig

JoerivanEngelen (changes requested):

PR #89, opened 2022-08-25
Document toml file

Hofer-Julian:


docs/configuration.rst
```diff
@@ -14,22 +14,149 @@ interpret a rawstring (``r''`` or ``r""``), whereas double quotes
 python (``""`` or ``''``). This matters for paths on Windows, for which
 we advice to use single quotes.
 
-.. code:: toml
+.. literalinclude:: config/metamod.toml
+    :language: toml
 
-   # This is a configuration file for the imod_coupler
-   # Relative paths are interpreted as relative to the configuration file path
+Config schema
```
JoerivanEngelen, line 20: Could you add for each variable what it roughly does? I think for most users names like ``log_level`` and ``mf6_msw_recharge_pkg`` are not self-explanatory. For example, the explanation for ``mf6_msw_recharge_pkg``  could be: "_Package name (specified in the Modflow 6 .nam file) of the recharge package to which MetaSWAP will be coupled_".

JoerivanEngelen (changes requested):
I like the tables with the description of types and possible values for each keyword in the toml file. 

I do however think that our users require a description of what all keywords stand for.

docs/configuration.rst
```diff
@@ -14,22 +14,142 @@ interpret a rawstring (``r''`` or ``r""``), whereas double quotes
...
+    type, string
+    required, true
+
+coupling
+^^^^^^^^
+
+.. csv-table:: enable_sprinkling
+    :widths: 3, 7
+
+    description, "Whether to enable_sprinkling."
```
JoerivanEngelen, line 110:
```suggestion
    description, "Whether to enable sprinkling, that is: enable extracting groundwater for irrigation."
```

JoerivanEngelen (commented):
Got one suggestion to make the description more specific. Since MetaSWAP also has an option to use surface water to irrigate crops.

PR #92, opened 2022-08-26
Add test and test cases for create_mapping function

JoerivanEngelen:


tests/test_mapping.py
```diff
@@ -0,0 +1,27 @@
...
+@parametrize_with_cases(
+    "src_idx,tgt_idx,nsrc,ntgt,operator,expected_map_dense,expected_mask"
+)
+def test_create_mapping(
+    src_idx, tgt_idx, nsrc, ntgt, operator, expected_map_dense, expected_mask
+):
+    map_out, mask = create_mapping(src_idx, tgt_idx, nsrc, ntgt, operator)
+
+    assert issubclass(map_out.dtype.type, np.floating)
+    assert issubclass(mask.dtype.type, np.integer)
```
Hofer-Julian, line 20: Theoretically we should be able to check that statically with mypy, but maybe that doesn't work yet relieably with numpy.

tests/test_mapping.py
```diff
@@ -0,0 +1,27 @@
...
+
+from imod_coupler.utils import create_mapping
+
+
+#%% Test
+@parametrize_with_cases(
+    "src_idx,tgt_idx,nsrc,ntgt,operator,expected_map_dense,expected_mask"
+)
+def test_create_mapping(
+    src_idx, tgt_idx, nsrc, ntgt, operator, expected_map_dense, expected_mask
```
Hofer-Julian, line 15: Many of the names here are not very obvious

tests/test_mapping_cases.py
```diff
@@ -0,0 +1,165 @@
+import numpy as np
+
+
+def case_1_1_symmetric_sum():
+    src_idx = np.array([0, 1, 2], dtype=int)
+    tgt_idx = np.array([0, 1, 2], dtype=int)
+
+    nsrc: int = 3
+    ntgt: int = 3
```
Hofer-Julian, line 9: Is that type hint necessary?

Hofer-Julian (commented):
Only small comments, maybe no changes are necessary.

tests/test_mapping_cases.py
JoerivanEngelen: Nope, removed!

tests/test_mapping.py
JoerivanEngelen: The argument names are equivalent to the `create_mapping` function, I'll write some docstring pointing to the `create_mapping` function.

Hofer-Julian: One more reason to refactor that function :P

PR #96, opened 2022-10-19
IMOD6-934: changes in create mapping to support weight average exchange and add …

HendrikKok:
…example lumbricus-driver


luitjansl (approved):

Hofer-Julian (changes requested):
Also here mypy errors need to be fixed before merging.

HendrikKok:
Superseded by https://github.com/Deltares/imod_coupler/pull/99

PR #98, opened 2022-10-21
IMOD6-944: Add initial version of DfmMetaMod driver

Hofer-Julian:


imod_coupler/drivers/match_driver.py
```diff
@@ -0,0 +1,50 @@
...
+from imod_coupler.config import BaseConfig
+from imod_coupler.drivers.dfm_metamod.config import DfmMetaModConfig
+from imod_coupler.drivers.dfm_metamod.dfm_metamod import DfmMetaMod
+from imod_coupler.drivers.driver import Driver
+from imod_coupler.drivers.metamod.config import MetaModConfig
+from imod_coupler.drivers.metamod.metamod import MetaMod
+
+
+def match_driver(
+    driver_type: str,
```
luitjansl, line 15: Relatively unimportant, but this function could be made more future proof by doing the following:
-add a name attribute to the driver base class
-assign different values to this attribute in the metamod and dfm_metamod classes
-loop here over the derived typed of Driver until one is found with the desired name
-call its constructor and return the object

That way, future new additions of drivers will not make it necessary to alter this function

luitjansl (approved):
looks good! I added one suggestion you can look at.

imod_coupler/drivers/match_driver.py
Hofer-Julian: Sounds good.
I've adapted the code accordingly.

PR #99, opened 2022-10-24
IMOD6-934: Dflowmetamod driver test

HendrikKok:
I put my Lumbricus-work for the exchange mapping between MF6-MetaSWAP-DFLOW-FM in the example-folder. This contains the following data:

1. input_example_mapping-folder contains the mapping files, which are now produced by imod.
2. mapping_functions.py contains alle the functions needed for the exchange-mappng between the kernels.
3. mapping_example.py is a testcase/test for the MF-DFLOW 1D exchange.


examples/mapping_example.py
```diff
@@ -0,0 +1,113 @@
...
+
+import numpy as np
+from mapping_functions import *
+
+
+def calculated_as_expected(expected, calculated):
+    ok = []
+    for i in range(len(expected)):
+        ok.append(isclose(expected[i], calculated[i], rel_tol=0.001, abs_tol=0.0))
+    return all(ok)
```
luitjansl, line 13: is it necessary to store the array "ok"?  from the looks of it, you could do:
for i in range(len(expected)):
        if not isclose(expected[i], calculated[i], rel_tol=0.001, abs_tol=0.0):
                 return False
return True

examples/mapping_example.py
```diff
@@ -0,0 +1,113 @@
...
+# create dummy arrays to exchange
+
+# mf riv1-flux to exchange
+mf_riv1_flux = np.array([3, 3, 4, 4, 4])
+# dflow1d flux and stage to exchange
+dflow1d_flux = np.array([6, 7, 8])
+dflow1d_stage = np.array([4, 5, 6])
+
+# get dflow-id based on xy-coordinates after initialisation (now as test from file)
+dflow1d_lookup, ok_file = get_dflow1d_lookup(workdir)
```
luitjansl, line 28: if ok_file is not used, you can make that explicit by renaming this variable to "_".
The underscore is commonly used for unused variables.

examples/mapping_example.py
```diff
@@ -0,0 +1,113 @@
...
+        (0.1 * 5) + (0.9 * 6),
+    ]
+)
+mf_riv1_stage_receive = np.array([0, 0, 0, 0, 0])
+mf_riv1_stage_receive = (
+    mask_active_mod_dflow1d["dflow1d2mf-riv_stage"][:] * mf_riv1_stage_receive[:]
+    + map_active_mod_dflow1d["dflow1d2mf-riv_stage"].dot(dflow1d_stage)[:]
+)
+if not calculated_as_expected(mf_riv1_stage_receive_expected, mf_riv1_stage_receive):
+    print("FOUT in exchange stage DFLOW 1D -> MF")
```
luitjansl, line 70: here you would typically raise an exception. Otherwise the program just continues.

imod_coupler/utils.py
```diff
@@ -13,7 +13,7 @@
 
 
 def create_mapping(
-    src_idx: Any, tgt_idx: Any, nsrc: int, ntgt: int, operator: str
+    src_idx: Any, tgt_idx: Any, nsrc: int, ntgt: int, operator: str, weight: Optional[NDArray] = None
```
luitjansl, line 16: for the operator string  you could use an enum:
from enum import Enum
 class Operation(Enum):
      Sum = auto()
      Average = auto()
      Weigths = auto()

examples/mapping_functions.py
```diff
@@ -0,0 +1,445 @@
...
+
+    # MF RIV 1 -> DFLOW 1D (flux)
+    mapping_file = workdir / "MFRIVTODFM1D_Q.DMM"
+    if mapping_file.is_file():
+        table_active_mfriv2dflow1d: NDArray[np.single] = np.loadtxt(
+            mapping_file, dtype=np.single, ndmin=2, skiprows=1
+        )
+        mf_idx = table_active_mfriv2dflow1d[:, 2].astype(int) - 1
+        dflow_idx = np.array(
+            [
```
luitjansl, line 35: I think here you don't need to go back to the ii index. You  might be able to fill this list as:

   [dflow1d_lookup[row[0], row[1]  for row in table_active_mfriv2dflow1d)   ]

luitjansl (commented):
nice work! I added some (random) comments about the use of python you may want to look at. 
In general, try to swap as little as possible between lists and numpy arrays. I saw that semtimes you convert from one type to the other to do some operation on the data. For here that is fine, but in the production code we should use numpy operations as much as possible.

examples/mapping_example.py
HendrikKok: it is not indeed, changed it into your suggestion

examples/mapping_example.py
HendrikKok: I changed those indeed

imod_coupler/utils.py
HendrikKok: this Optional specification was added by Julian. Can i keep it like this?

Hofer-Julian: Luit Jan meant the operator, not the weight.
I agree, but would suggest [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum) instead

Hofer-Julian: Correction: Luit-Jan's suggestion is better.

Hofer-Julian:
Also mypy errors need to be fixed.
Maybe @luitjansl can help with that.

imod_coupler/utils.py
HendrikKok: Done

Hofer-Julian (approved):

PR #103, opened 2022-11-04
Imod6 939 couple mf and dfm

luitjansl:
-contains the setup for exchanging fluxes in the timeloop 
-the exchange fluxes function names are based on the names these fluxes have in the numerical scheme picture
-the mapping is not yet used, and the dflow example data does not have rivers
-unittests were made, but some hade to be skipped due to a suspected bug in either DFlowFM or its BMI wrapper. For some reason, we cannot initialize a wrapper once another one was initialized and finalized in the same session
-these unittests should work when executed individually though


imod_coupler/drivers/dfm_metamod/mf6_utilities.py
```diff
@@ -0,0 +1,27 @@
+import numpy as np
+from imod import mf6
```
Hofer-Julian, line 2: You cannot depend on imod-python :)

Hofer-Julian: (in the main code at least, the tests are fine)

luitjansl: done. removed these imports. They were unused.

Hofer-Julian: Nice, now the testbench is green too

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
```diff
@@ -38,30 +40,14 @@ class DfmMetaMod(Driver):
...
-    map_msw2mod: Dict[str, csr_matrix] = {}
-    # dict. with mask arrays for mod=>msw coupling
-    mask_mod2msw: Dict[str, NDArray[Any]] = {}
-    # dict. with mask arrays for msw=>mod coupling
-    mask_msw2mod: Dict[str, NDArray[Any]] = {}
+    mf6_head: NDArray[Any]  # the hydraulic head array in the groundwater model
+    mf6_river_stage: NDArray[Any]  # the river stage array in the groundwater model
+
+    dflowfm_1d_stage: NDArray[
+        Any
```
Hofer-Julian, line 47: This `Any` is still a relic from before I know how to properly use typing with numpy.
You should, depending on whether it is a float or integer, use `np.float_` or `np.int_` instead

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
```diff
@@ -77,13 +63,17 @@ def initialize(self) -> None:
         self.mf6 = XmiWrapper(
             lib_path=self.dfm_metamod_config.kernels.modflow6.dll,
             lib_dependency=self.dfm_metamod_config.kernels.modflow6.dll_dep_dir,
-            working_directory=self.dfm_metamod_config.kernels.modflow6.work_dir,
+            working_directory=Path(
+                self.dfm_metamod_config.kernels.modflow6.work_dir.absolute()
```
Hofer-Julian, line 67: The `absolute` should not be necessary, we take already care for that in the `config` module by calling `resolve`

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
```diff
@@ -77,13 +63,17 @@ def initialize(self) -> None:
...
+            ),
             timing=self.base_config.timing,
         )
         self.msw = XmiWrapper(
             lib_path=self.dfm_metamod_config.kernels.metaswap.dll,
             lib_dependency=self.dfm_metamod_config.kernels.metaswap.dll_dep_dir,
-            working_directory=self.dfm_metamod_config.kernels.metaswap.work_dir,
+            working_directory=Path(
+                self.dfm_metamod_config.kernels.metaswap.work_dir.absolute()
```
Hofer-Julian, line 75: Same here

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
```diff
@@ -118,205 +109,74 @@ def couple(self) -> None:
...
-            has_converged = self.do_iter(1)
-            if has_converged:
-                logger.debug(f"MF6-MSW converged in {kiter} iterations")
-                break
-        self.mf6.finalize_solve(1)
+        self.exchange_H_1D_t()
+        self.exchange_V_1D()
+
+        subtimestep_endtime = self.get_current_time()
+        for _ in range(0, self.number_dflowsteps_per_modflowstep):
```
Hofer-Julian, line 125:
```suggestion
        for _ in range(self.number_dflowsteps_per_modflowstep):
```

imod_coupler/drivers/dfm_metamod/dfm_utilities.py
```diff
@@ -0,0 +1,27 @@
+import numpy as np
+from bmi.wrapper import BMIWrapper
+from numpy.typing import NDArray
+
+
+class DfmUtilities:
+    @classmethod
```
Hofer-Julian, line 7: As discussed before, instead of adding a couple of `classmethod`s that all take the dflow `BMIWrapper`, subclass `BMIWrapper` instead. `mypy` will complain since `BMIWrapper` is untyped, but you can just ignore this as shown here: https://github.com/Deltares/imod_coupler/blob/1dcb67bb252911cb107f95655f75ac84c1320188/imod_coupler/drivers/dfm_metamod/dfm_wrapper.py#L6

imod_coupler/drivers/dfm_metamod/dfm_utilities.py
```diff
@@ -0,0 +1,27 @@
...
+    def get_number_1d_nodes(cls, dflow: BMIWrapper) -> int:
+        nr_nodes = dflow.get_var("ndxi")  # number of 1d cells
+        nr_nodes2d = dflow.get_var("ndx2d")  # number of 2d cells
+        return int(nr_nodes - nr_nodes2d)
+
+    @classmethod
+    def get_waterlevels_1d(cls, dflow: BMIWrapper) -> NDArray[np.float_]:
+        nr_nodes_1d = cls.get_number_1d_nodes(dflow)
+        if nr_nodes_1d == 0:
+            return np.empty(shape=(0), dtype=np.float_)
```
Hofer-Julian, line 17: Why not return `None` for this case?

imod_coupler/drivers/dfm_metamod/dfm_utilities.py
```diff
@@ -0,0 +1,27 @@
...
+        all_waterlevels = dflow.get_var("s1")
+        return np.asarray(all_waterlevels[-nr_nodes_1d:], dtype=np.float_)
+
+    @classmethod
+    def get_cumulative_fluxes_1d_nodes(cls, dflow: BMIWrapper) -> NDArray[np.float_]:
+        nr_nodes_1d = cls.get_number_1d_nodes(dflow)
+        if nr_nodes_1d == 0:
+            return np.empty(shape=(0), dtype=np.float_)
```
Hofer-Julian, line 25: Same here

imod_coupler/drivers/dfm_metamod/mf6_utilities.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+from numpy.typing import NDArray
+from xmipy import XmiWrapper
+
+
+class MF6Utilities:
+    @classmethod
+    def set_river_stages(
```
Hofer-Julian, line 8: I would also subclass `XmiWrapper` here

imod_coupler/drivers/dfm_metamod/mf6_utilities.py
```diff
@@ -0,0 +1,25 @@
...
+        cls,
+        mf6_wrapper: XmiWrapper,
+        mf6_flowmodel_key: str,
+        mf6_river_pkg_key: str,
+        new_river_stages: NDArray[np.float_],
+    ) -> None:
+        """
+        sets the river stages in a modflow simulation to the provided values.
```
Hofer-Julian, line 16: That is useful for generating python docstrings: https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring

We currently use the numpy conventions: https://numpydoc.readthedocs.io/en/v1.1.0/format.html

tests/test_mf6_utilities.py
```diff
@@ -0,0 +1,40 @@
...
+        "GWF_1",
+        "Oosterschelde",
+        new_river_stage,
+    )
+
+    bound_adress = mf6wrapper.get_var_address("BOUND", "GWF_1", "Oosterschelde")
+    bound = mf6wrapper.get_value_ptr(bound_adress)
+    stage = bound[:, 0]
+
+    assert np.isclose(stage, new_river_stage).all()
```
Hofer-Julian, line 39:
```suggestion
    np.testing.assert_allclose(stage, new_river_stage)
```

Hofer-Julian (approved):
I've added a couple of minor comments.
After these are fixed, the PR is good to be merged from my side :)

imod_coupler/drivers/dfm_metamod/mf6_utilities.py
Hofer-Julian: Here I would have expected that the parameters are documented.

imod_coupler/drivers/dfm_metamod/dfm_utilities.py
luitjansl: done

imod_coupler/drivers/dfm_metamod/mf6_utilities.py
luitjansl: done

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
```diff
@@ -33,35 +34,19 @@ class DfmMetaMod(Driver):
     timing: bool  # true, when timing is enabled
     mf6: XmiWrapper  # the MODFLOW 6 XMI kernel
```
Hofer-Julian, line 35: Also remember to change this type :)

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
luitjansl: done

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
luitjansl: not done. If I drop the ". absolute()", or if I also drop the path constructor and just use 
`working_directolry =self.dfm_metamod_config.kernels.modflow6.work_dir `
Then after the call to "self.dfm.initialize()" the modflow working directory is miraculously changed from XXX\MF6 to XXX\dflow_fm\MF6
This is maybe due to the fact that Python uses so much references. By assigning a deep copy of the path this effect is avoided.

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
luitjansl: see comment above.

imod_coupler/drivers/dfm_metamod/dfm_utilities.py
luitjansl: because of mypy

imod_coupler/drivers/dfm_metamod/dfm_utilities.py
luitjansl: because of mypy

imod_coupler/drivers/dfm_metamod/mf6_utilities.py
luitjansl: done. Installed the plugin and set settings to numpy format.

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
Hofer-Julian: > By assigning a deep copy of the path this effect is avoided.

In that case I would make explicitly a deepcopy to make the intention clear.

imod_coupler/drivers/dfm_metamod/dfm_utilities.py
Hofer-Julian: You can make the return type `Optional[NDArray[np.float_]]`.

Import it from `from typing import Optional`

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
luitjansl: done

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
luitjansl: done

imod_coupler/drivers/dfm_metamod/dfm_utilities.py
luitjansl: done

imod_coupler/drivers/dfm_metamod/dfm_utilities.py
luitjansl: done

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
luitjansl: done. Now using 
`            working_directory=copy.deepcopy(
                self.dfm_metamod_config.kernels.modflow6.work_dir.absolute()
            )`
I tired dropping the .absolute() as well, but then I still had the problem outlined above. baffling.

PR #109, opened 2022-11-14
Fix incorrect initializition of class variables

Hofer-Julian:


imod_coupler/drivers/metamod/metamod.py
```diff
@@ -122,6 +122,11 @@ def couple(self) -> None:
         self.msw_volume = self.msw.get_value_ptr("dvsim")
         self.msw_storage = self.msw.get_value_ptr("dsc1sim")
 
+        self.map_mod2msw = {}
```
luitjansl, line 125: I think here we can do: self.map_mod2msw= Dicr[str, NDArray[Any]]. Then the definition above (line 53) can be removed altogether. I believe this communicates the intent of the programmer better, because then you can access (the object variable) mydriver.map_mod2msw, but you cannot access (the class variable) metamod.map_mod2msw.

luitjansl (approved):

imod_coupler/drivers/metamod/metamod.py
Hofer-Julian: Both should have the exact same behaviour (if you can find differences please send a small test script).

Personally I like type annotations on the class variable level more and I also see it used more often.

See more here: https://mypy.readthedocs.io/en/stable/class_basics.html#instance-and-class-attributes

PR #110, opened 2022-11-15
iMOD6-961 add MetaMOD functionality to our driver

HendrikKok:
I updated driver with metamod-functionality


imod_coupler/drivers/dfm_metamod/dfm_metamod.py
```diff
@@ -63,6 +77,15 @@ class DfmMetaMod(Driver):
     # masks used for  modflow-dflow exchanges
     mask_active_mod_dflow1d = dict[str, NDArray[np.int_]]()
 
+    # dictionary with mapping tables for mod=>msw coupling
+    map_mod2msw: Dict[str, csr_matrix] = {}
```
Hofer-Julian, line 81: Initializing with a dict at that level turned out to be problematic.
Please change that similar to what I did in the last commit of lumbricus: https://github.com/Deltares/imod_coupler/commit/a798c9c8071cf7f0fe03c96bce48013b1a906abe

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
```diff
@@ -127,28 +149,157 @@ def couple(self) -> None:
...
+
+        self.map_msw2mod["recharge"], self.mask_msw2mod["recharge"] = create_mapping(
+            msw_idx,
+            rch_idx,
+            self.msw_volume.size,
+            self.mf6_recharge.size,
+            Operator.SUM,
+        )
+
+        if self.coupling.enable_sprinkling:
```
luitjansl, line 238: since we decided to skip sprinkling for now, please remove everything that has to do with sprinkling: teh pointers, the mappings, etc

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
```diff
@@ -218,8 +369,56 @@ def exchange_V_dash_1D(self) -> None:
...
+            + self.map_mod2msw["head"].dot(self.mf6_head)[:]
+        )
+
     def report_timing_totals(self) -> None:
         total_mf6 = self.mf6.report_timing_totals()
         total_msw = self.msw.report_timing_totals()
         total = total_mf6 + total_msw
         logger.info(f"Total elapsed time in numerical kernels: {total:0.4f} seconds")
+
+    def do_iter(self, sol_id: int) -> bool:
```
luitjansl, line 416: do_iter_modflow_and_metaswap ?

luitjansl, line 416: I'm not sure what sol_id is, but it is called with the value "1" every time, so could it be a local variable instead of a function parameter?

luitjansl (approved):

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
```diff
@@ -63,6 +77,15 @@ class DfmMetaMod(Driver):
...
     mask_active_mod_dflow1d = dict[str, NDArray[np.int_]]()
 
+    # dictionary with mapping tables for mod=>msw coupling
+    map_mod2msw: Dict[str, csr_matrix] = {}
+    # dictionary with mapping tables for msw=>mod coupling
+    map_msw2mod: Dict[str, csr_matrix] = {}
+    # dict. with mask arrays for mod=>msw coupling
+    mask_mod2msw: Dict[str, NDArray[Any]] = {}
+    # dict. with mask arrays for msw=>mod coupling
+    mask_msw2mod: Dict[str, NDArray[Any]] = {}
```
Hofer-Julian, line 87: Same for the other dicts

Hofer-Julian (approved):
Rest looks fine to me.
After the comments are addressed and we get TeamCity to be green it is fine for me to merge.

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
HendrikKok: ok, its changed. For my information, why did this give problems? I can see at least it gives cleaner code.

imod_coupler/drivers/dfm_metamod/mapping_functions.py
```diff
@@ -439,6 +439,38 @@ def get_dflow1d_lookup(workdir: Path) -> tuple[dict[tuple[float, float], int], b
...
+    if msw_mod2svat_file.is_file():
+        svat_data: NDArray[np.int32] = np.loadtxt(
+            msw_mod2svat_file, dtype=np.int32, ndmin=2
+        )
+        svat_id = svat_data[:, 1]
+        svat_lay = svat_data[:, 2]
+        for vi in range(svat_id.size):
+            svat_lookup[(svat_id[vi], svat_lay[vi])] = vi
+    else:
+        ok = False
```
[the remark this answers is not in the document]
HendrikKok: true, done!

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
HendrikKok: This is the sprinkling from groundwater and is part of the 'normal' coupling MSW-MF6. We skip only the sprinkling form surface water right?

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
HendrikKok: done

HendrikKok: yes, i was asking myself the same question when copying it. If you look at the underlying solve in the xmiwrapper.py, it seems that it is actually a local variable right   --> solve(self, component_id=1)

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
HendrikKok: done

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
Hofer-Julian: The problem is that the former code resulted in class variable set to a single dictionary.
Imagine the following situation with two different instances of `DfmMetaMod`:
```python
>>> driver_1 = DfmMetaMod()
>>> driver_2 = DfmMetaMod()
>>> driver_1.map_mod2msw["blah"] = "whooo"
>>> print(driver_2.map_mod2msw["blah"])
whooo
```

Not a situation that will happen very often, but still a bug.
And it is only a problem if you set the default value of a class variable to a mutable container like dict or a numpy array.

PR #114, opened 2022-12-05
add wel package to t-model for correction flux (name = RIV_CORR)

HendrikKok:


luitjansl:
I think this pull request can be closed, because the changes in it were committed in commit d4fac75a2c0edf3aa5622c78717c4f6fde09927d

Hofer-Julian:
@HendrikKok do you agree?

PR #115, opened 2022-12-06
Imod6 976 correction flux

luitjansl:
-the functions that compute or set the correction flux were already made, but they were not used in the program
-the well package used for the correction flux in the modflow model was made by Hendrik, but it was not present in our t-model

so

-this changeset takes these building blocks and calls them from the driver to implement the correction flux
-also, a unit conversion was added. dflow-fm reports its time in seconds, while mf6 uses days  ( in our example)


imod_coupler/drivers/dfm_metamod/dfm_metamod.py
```diff
@@ -289,14 +291,24 @@ def update(self) -> None:
         # flux from modflow to dflow 1d
         self.exchange_V_1D()
 
+        q_corr_init = self.dfm.get_cumulative_fluxes_1d_nodes()
+        assert q_corr_init is not None
+
         # sub timestepping between metaswap and dflow
-        subtimestep_endtime = self.get_current_time()
+        subtimestep_endtime = tBegin
```
Hofer-Julian, line 298:
```suggestion
        subtimestep_endtime = t_begin
```

Same for other occurrences of `tBegin`

imod_coupler/drivers/dfm_metamod/mf6_wrapper.py
```diff
@@ -135,7 +135,10 @@ def set_correction_flux(
         """
         bound_adress = self.get_var_address("BOUND", mf6_flowmodel_key, mf6_wel_pkg_key)
         flux = self.get_value_ptr(bound_adress)
+
         if correction_flux is None or len(correction_flux) != len(flux):
             raise ValueError(f"Expected size of correction_flux is {len(flux)}")
-        flux[:] = correction_flux[:]
+        for i in range(len(flux)):
+            flux[i, 0] = correction_flux[i]
```
Hofer-Julian, line 142: Where does this second dimension of `flux` come from?

Hofer-Julian (approved):
Only small comments.

It seems like TeamCity timed out, I've triggered another run, let's see if that one will work out.
Are the tests working for you locally?

Hofer-Julian:
Locally, it fails for me with:
```
FAILED tests/test_run_tmodel.py::test_run_tmodel - pydantic.error_wrappers.ValidationError: 1 validation error for DfmMetaModConfig
```
Not sure why it times out on TeamCity.

imod_coupler/drivers/dfm_metamod/dfm_metamod.py
luitjansl: done

luitjansl:
>Not sure why it times out on TeamCity.
i found out it times out because of a precision issue.             
in the statement:
while  self.dfm.get_current_time() < days_to_seconds(subtimestep_endtime  ):

it so happens that for the final time 365 days,  days_to_seconds(subtimestep_endtime  ) returns 31536000.0000002 whereas dflow-fm can compute up to 31536000. So no matter how many extra dfm timesteps we make, the ending condition is never met. I now added a tolerance to this comparison, and now it finishes.

PR #121, opened 2023-01-16
Fix imod Python commit

Hofer-Julian:
That keeps the testbench green for now

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,7 +15,7 @@ dependencies:
   - pip
   - pip:
       - xmipy
-      - git+https://gitlab.com/deltares/imod/imod-python.git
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
See https://issuetracker.deltares.nl/browse/IMOD6-985

Hofer-Julian:
> so you're collecting a specific version of imod that works with our tests, is that it? If that's the case, then it looks good to me. (can you confirm that that?)

Yes, indeed. Only until the underlying problem is solved after discussion with @JoerivanEngelen.

> Maybe imod-python needs explicit version numbers to deal with these things (so we an say that the coupler works with imod-python version 1 .0.3 instead of using the githash) but that's another discussion.

It has them, but so far we needed a newer setup than the latest release.
I would be happy to just depend on the latest release, which should be possible after the next imod python.

luitjansl (approved):

PR #122, opened 2023-01-24
Imod6 990 test riv2 mapping

luitjansl:
added a test for riv2 and drn2 mappings and masks calculation


tests/test_mapping_example.py
```diff
@@ -173,22 +173,25 @@ def test_mapping_riv2_drain(
...
-            [0.0, 0.0],
-            [0.0, 0.0],
-            [0.0, 0.0],
-            [0.0, 0.0],
-            [0.0, 0.0],
-            [1.0, 0.0],
-            [0.0, 1.0],
-        ]
-    )
+    assert (
```
Hofer-Julian, line 176: This one has a nicer error message: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html

You can get a ndarray by using toarray instead of todense

luitjansl: done

PR #125, opened 2023-01-27
No convergence in MODFLOW doesn't result in an error anymore

Hofer-Julian:
See https://github.com/MODFLOW-USGS/modflow6/pull/1141 by @mjr-deltares 

@luitjansl You will probably want to merge this change into the lumbricus branch

### DIFF
--- a/tests/test_metamod_cases.py
+++ b/tests/test_metamod_cases.py
@@ -133,28 +133,6 @@ def fail_run_mf6_input(
     )
 
 
-def fail_run_mf6_convergence(
-    coupled_mf6_model: Modflow6Simulation,
-    prepared_msw_model: MetaSwapModel,
-) -> MetaMod:
-    """
-    Creates a MetaMod object which will result in a non-convergencent MODFLOW 6
-    solution, by providing extreme differences in the k values.
-    """
-
-    k = xr.ones_like(coupled_mf6_model["GWF_1"]["npf"]["k"])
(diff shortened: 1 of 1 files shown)


luitjansl (approved):
looks good to me. If I understood correctly, Martijn will still need to update xmipy to make that   survive if mf6 does not converge. Has that already happened? You could consider waiting for that before removing this testcase. After all, this testcase documents the current expected behavior. Anyway, just a thought.

Hofer-Julian:
> . If I understood correctly, Martijn will still need to update xmipy to make that survive if mf6 does not converge.

As far as I can tell, it already survives it now.
They way I've understood is that the change could happen in MODFLOW or xmipy, but MODFLOW is the better choice.
MODFLOW code has now been adapted.

mjr-deltares:
> > . If I understood correctly, Martijn will still need to update xmipy to make that survive if mf6 does not converge.
> 
> As far as I can tell, it already survives it now. They way I've understood is that the change could happen in MODFLOW or xmipy, but MODFLOW is the better choice. MODFLOW code has now been adapted.

Yes, that is the intended behavior. If it doesn't work out this way, please let me know.

Hofer-Julian:
@mjr-deltares The test was assuming that no convergence leads to a bmi error.
That assumption does not hold true anymore, so I would say your change works as expected :)

PR #127, opened 2023-01-31
I mod6 998 extend msw wrapper

HendrikKok:
created and extended msw-wrapper with all MetaSWAP bmi/xmi-calls we need for the coupling of MetaSWAP-DFLOW. For a couple of functions, the BMI of MetaSWAP need to be extended first.


imod_coupler/drivers/dfm_metamod/msw_wrapper.py
```diff
@@ -0,0 +1,122 @@
...
+
+        """
+        function to initialise the surface water component in metaswap.
+
+        Parameters
+        ----------
+        none
+
+        """
+        self.initiate_surface_water_component()
```
luitjansl, line 19: sonarqube has a comment about this function: it calls itself and will get stuck in an infinite loop.

imod_coupler/drivers/dfm_metamod/msw_wrapper.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        function to start and excecute surface water timestep between msw and dflow
+
+        Parameters
+        ----------
+        idtsw : integer*4
+            index of time step within dtgw-cycle
+
+        """
+        self.start_surface_water_time_step(idtsw)
```
luitjansl, line 31: same here

imod_coupler/drivers/dfm_metamod/msw_wrapper.py
```diff
@@ -0,0 +1,122 @@
...
+        Returns
+        -------
+         NDArray[np.float_]:
+            ponding volume allocation of MetaSWAP in m3/dtsw. Array as a copy of the MetaSWAP intenal array,
+            since the set function uses a different bmi/xmi-variable
+        """
+        ponding_allocation = self.get_value("ts2dfmput")
+        return ponding_allocation
+
+    def put_surfacewater_ponding_allocation(
```
luitjansl, line 94: " set" is more standard than " put"  so maybe "set_surfacewater_ponding_allocation"?

imod_coupler/drivers/dfm_metamod/msw_wrapper.py
```diff
@@ -0,0 +1,122 @@
...
+            ponding volume allocation to set in metaswap in m3/dtsw
+
+        Returns
+        -------
+        none
+        """
+        self.set_value("ts2dfmget", ponding_allocation)
+
+    def put_ponding_level_2d(self, ponding_level_2d: NDArray[np.float_]) -> None:
+        """sets ponding level from dlfow-2d in metaswap
```
luitjansl, line 111: since we have a "get_ponding_level_2d" and a "put_ponding_level_2d", can we make a test that calls the set first and then the get, and checks if it returns the same?

HendrikKok: that's not possible unfortunately. Put and set refer to a different variable within MetaSWAP. Hence the separate functions

luitjansl (approved):

PR #128, opened 2023-02-03
deleted variable msw_time

HendrikKok:
msw_time only exist within the MetaMod-driver and is not used by kernels.


Hofer-Julian (changes requested):
You need the changes in `main` for the CI to be green.
Run:
```
git switch delete_msw_time
git fetch origin
git rebase main
git push --force-with-lease
```

That way the branch should be updated, the CI green and we can merge this.

Hofer-Julian (approved):
Thanks Hendrik!

PR #129, opened 2023-02-03
Imod6 999 refactoring metamod

HendrikKok:
Refactoring Metamod-functionality within our Dfm-metamod driver. Couple function of Metamod moved to MF6 and MSW wrappers and to the new mapping class.


imod_coupler/drivers/dfm_metamod/mf6_wrapper.py
```diff
@@ -1,26 +1,92 @@
...
+from pathlib import Path
+from typing import Any, Optional, Union
 
 import numpy as np
 from numpy.typing import NDArray
 from xmipy import XmiWrapper
 
 
 class Mf6Wrapper(XmiWrapper):
+    def __init__(
```
luitjansl, line 10: as discussed offline: some changes in this file should be reverted. We now pass the name of the flow-model and package-keys in the constructor. But that is limiting the general usefulness of this wrapper. These should be passed as function arguments, so that the users of this wrapper have freedom to get data out of any package they want-and not necesarily the ones they gave to the constructor. Either create a new task for that or fix it here.

imod_coupler/drivers/dfm_metamod/msw_wrapper.py
```diff
@@ -0,0 +1,213 @@
...
+
+        """
+        function to initialise the surface water component in metaswap.
+
+        Parameters
+        ----------
+        none
+
+        """
+        idtsw = 0
```
luitjansl, line 29: this function creates 2 local variables, sets them to 0, and then it is finished? that's what it looks like anyway. then it could be removed!

tests/test_mapping_example.py
```diff
@@ -15,6 +17,10 @@
 from imod_coupler.utils import Operator, create_mapping
 
 
+def create_mapper():
+    return Mapping()
```
luitjansl, line 21: is this used?

imod_coupler/drivers/dfm_metamod/mf6_wrapper.py
```diff
@@ -1,26 +1,92 @@
...
+        return mf6_head
+
+    def recharge(self) -> NDArray[np.float_]:
+        mf6_recharge_tag = self.get_var_address(
+            "BOUND", self.mf6_flowmodel_key, self.mf6_msw_recharge_pkg
+        )
+        mf6_recharge = self.get_value_ptr(mf6_recharge_tag)[:, 0]
+        return mf6_recharge
+
+    def storage(self) -> NDArray[np.float_]:
```
luitjansl, line 39: head, recharge and storage and others are getters. Please call them get_*

imod_coupler/drivers/dfm_metamod/mapping.py
```diff
@@ -0,0 +1,622 @@
...
+import numpy as np
+from numpy import float_, int_
+from numpy.typing import NDArray
+from scipy.sparse import csr_matrix, dia_matrix, diags
+
+from imod_coupler.utils import Operator, create_mapping
+
+
+class Mapping:
+    def __init__(self, coupling: Any, mf6_wrapper: Any, msw_wrapper: Any) -> None:
```
luitjansl, line 15: as discussed, we need to think about making this class independent of these constructor arguments, because now it has become hard to set up tests for this class (they requiere a coupling object, a mf6_wrapper etc) 
In any case for now please provide more specific types (not " Any" ) for the mypy type checking. If everything is "Any" then type checking becomes not so usefull.

luitjansl (commented):
It's a very nice cleanup, but I also have some comments. Either fix them here or create follow-up issues to fix them later.

PR #130, opened 2023-02-03
-merged from main

luitjansl:
-adapted river model to pass imod-python validation

### DIFF
--- a/tests/test_metamod.py
+++ b/tests/test_metamod.py
@@ -44,6 +44,29 @@ def test_metamod_write_failure(
         )
 
 
+@parametrize_with_cases("metamod_model", prefix="fail_write_")
+def test_metamod_write_failure(
+    tmp_path_dev: Path,
+    metamod_model: MetaMod,
+    metaswap_dll_devel: Path,
+    metaswap_dll_dep_dir_devel: Path,
+    modflow_dll_devel: Path,
+) -> None:
+    """
+    Test if iMOD Python throws a error during writing coupled models that
+    shouldn't work.
+    """
+
+    # FUTURE: This might change to ValidationError
+    with pytest.raises(ValueError):
+        metamod_model.write(
+            tmp_path_dev,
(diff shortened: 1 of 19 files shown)


Hofer-Julian (approved):

tests/test_metamod.py
```diff
@@ -21,6 +21,52 @@ def test_lookup_table_present(metaswap_lookup_table: Path) -> None:
...
+    expected_dependencies = (
+        "fmpich2.dll",
+        "mpich2mpi.dll",
+        "mpich2nemesis.dll",
+        "TRANSOL.dll",
+    )
+
+    for dependency in expected_dependencies:
+        assert dependency in dep_dir_content
+
```
Hofer-Julian, line 68: I think these tests have been moved, so you don't need them here again.

tests/test_modstrip.py
```diff
@@ -49,6 +47,20 @@ def write_toml(
...
+    Fill para_sim.inp template in the folder with the path to the unsaturated
+    zone database.
+    """
+    with open(msw_folder / "para_sim_template.inp") as f:
+        para_sim_text = f.read()
+
+    para_sim_text = para_sim_text.replace("{{unsat_path}}", f"{path_unsat_dbase}\\")
+
+    with open(msw_folder / "para_sim.inp", mode="w") as f:
+        f.write(para_sim_text)
```
Hofer-Julian, line 61: Same here

Hofer-Julian (changes requested):

tests/test_metamod.py
luitjansl: you're right- I removed them now (on main they are still here- we only moved them on the lumbricus branch)

tests/test_modstrip.py
luitjansl: done

Hofer-Julian (approved):
Thank you, @luitjansl!

PR #132, opened 2023-02-08
Imod6 991 add drn riv2 after refactor

HendrikKok:


luitjansl:
this is a new version of pull request Imod6 991 add drn riv2; did you check the review comments I had on that?

imod_coupler/drivers/dfm_metamod/mf6_wrapper.py
```diff
@@ -122,16 +122,55 @@ def get_river_stages(
         stage = bound[:, 0]
         return stage
 
-    def get_river_flux(
+    def set_correction_flux(
```
luitjansl, line 125: from a modflow perspective, this function sets the pumping rate of the wells in a well package. So I think the function should be called  "set_pumping_rate" , not " set_correction_flux" d

imod_coupler/drivers/dfm_metamod/mf6_wrapper.py
```diff
@@ -122,16 +122,55 @@ def get_river_stages(
...
         self,
         mf6_flowmodel_key: str,
         mf6_package_key: str,
```
luitjansl, line 128: can you add a test for this function in the test_mf6_wrapper  file?

imod_coupler/drivers/dfm_metamod/mf6_wrapper.py
HendrikKok: indeed

imod_coupler/drivers/dfm_metamod/mf6_wrapper.py
HendrikKok: done

luitjansl (approved):

PR #133, opened 2023-02-20
created test for dfm_metamod_model output

luitjansl:
added path to output configuration file to toml, and created a test


HendrikKok:
Looks nice, we should also add the mf6_drain_pkg key

rleander (commented):
Minor comments (but does not really regard the changes here):
constructor dfmMetaMod had a huge number of positional arguments, which is prone to mistakes not easily detected.
We should find another solution for it.
Let's avoid forward slashes in paths and use pathjoin to concat paths in future.

PR #137, opened 2023-02-23
Imod6 1002 output class

rleander:


tests/test_exchange_collector.py
```diff
@@ -0,0 +1,109 @@
...
+            "mf-ridv2dflow1d_flux_output", some_smaller_array, 23.0
+        )
+    assert (
+        "operands could not be broadcast together with remapped shapes [original->remapped]: (4,)  and requested shape (1,5)"
+        in str(e.value)
+    )
+    exchange_collector.finalize()
+
+
+def test_exchange_collector_raises_exception_when_time_is_repeated(
```
luitjansl, line 86: the test name does not appear to reflect the test content

tests/test_exchange_collector.py
```diff
@@ -0,0 +1,109 @@
...
+    exchange_collector.log_exchange("mf-ridv2dflow1d_flux_output", some_arrays[2], 10.0)
+    exchange_collector.log_exchange("mf-ridv2dflow1d_flux_output", some_arrays[3], 9.0)
+    exchange_collector.finalize()
+
+    ds = nc.Dataset("./mf-ridv2dflow1d_flux_output.nc", "r")
+    dat = ds.variables["xchg"][:]
+    tim = ds.variables["time"][:]
+    assert np.array_equal(dat[1, :], some_arrays[3], equal_nan=True)
+    assert np.array_equal(tim[:], np.array([8.0, 9.0, 10.0]), equal_nan=True)
+    pass
```
luitjansl, line 109: why is there a "pass" here?

luitjansl (commented):
looks good. I added a few comments

imod_coupler/drivers/dfm_metamod/dfm_metamod_model.py
```diff
@@ -233,6 +244,7 @@ def write_toml(
             "timing": False,
             "log_level": "INFO",
             "driver_type": "dfm_metamod",
+            "output"
```
Hofer-Julian, line 247: Why doesn't this key have a value?

tests/test_exchange_collector.py
rleander: You are right... should have been removed. Used that line to put a breakpoint

imod_coupler/drivers/dfm_metamod/dfm_metamod_model.py
luitjansl: removed this key

imod_coupler/drivers/dfm_metamod/exchange_collector.py
```diff
@@ -0,0 +1,87 @@
...
+            self.initfile(len(exchange))
+        loc = np.where(self.timevar[:] == time)
+        if np.size(loc) > 0:
+            self.datavar[loc[0], :] = exchange[:]
+        else:
+            self.timevar[self.pos] = time
+            self.datavar[self.pos, :] = exchange[:]
+            self.pos += 1
+        if sync:
+            self.ds.sync()
```
HendrikKok, line 50: What does this sync() option do exactly? we use sync:bool at end of a timestep?

imod_coupler/drivers/dfm_metamod/exchange_collector.py
```diff
@@ -0,0 +1,87 @@
...
+    def write_exchange(
+        self, exchange: NDArray[Any], time: float, sync: bool = False
+    ) -> None:
+        if len(self.ds.dimensions) == 0:
+            self.initfile(len(exchange))
+        loc = np.where(self.timevar[:] == time)
+        if np.size(loc) > 0:
+            self.datavar[loc[0], :] = exchange[:]
+        else:
+            self.timevar[self.pos] = time
```
HendrikKok, line 46: do we need additional check if time(i) > time(i-1)?

tests/fixtures/fixture_exchange_collector.py
```diff
@@ -0,0 +1,20 @@
...
+
+@pytest_cases.fixture(scope="function")
+def output_config_toml() -> str:
+    input_file_content = """
+                        [[general]]
+                        output_dir = "."
+
+                        [[exchanges]]
+                        
+                        [[exchanges.mf-ridv2dflow1d_flux_output]]
```
HendrikKok, line 12: How do we deal with writing errors in labels? the labels dont seem to correspond to the exchangelabels of toml file? Also we don't define "towards" and "from" fluxes in toml file.

tests/test_dfm_metamod_model.py
```diff
@@ -0,0 +1,89 @@
+import textwrap
```
HendrikKok, line 1: why is this new? it was already existing right?

HendrikKok (commented):
Nice work guys! look clean to me, I had a few questions.

tests/test_dfm_metamod_model.py
luitjansl: maybe it was used in other parts of the program, but the imports have to be present in every file that uses them. textwrap is used in this file (line 42), so it has to be imported.

imod_coupler/drivers/dfm_metamod/exchange_collector.py
luitjansl: this updates a netcdf file on disk with what was added to the logger.So (if i understand correctly) each timestep the netcdf file would be updated with the latest exchange fluxes, and then we don't have to keep everything in memory

imod_coupler/drivers/dfm_metamod/exchange_collector.py
luitjansl: it doesn't need to be. This can be seen in test test_exchange_collector_overwrites_when_time_is_repeated  where we 
add logging for time 8, then 9 , then 10, and then 9 again. 


    exchange_collector.log_exchange("example_stage_output", some_arrays[0], 8.0)
    exchange_collector.log_exchange("example_stage_output", some_arrays[1], 9.0)
    exchange_collector.log_exchange("example_stage_output", some_arrays[2], 10.0)
    exchange_collector.log_exchange("example_stage_output", some_arrays[3], 9.0)

PR #138, opened 2023-02-27
Imod6 1031 refactor balance script

luitjansl:
added mypy verification to balance script and created test based on csv output


luitjansl:
closed because branch needs cleanup

PR #141, opened 2023-03-01
Imod6 1039 investigate balance mismatch teamcity

luitjansl:


luitjansl:
this branch was used to force runs on teamcity and to ivestigate the erratic behavior on teamcity. This issue appears solved now that we do model runs as subprocesses, so this branch can be closed

PR #151, opened 2023-03-23
Imod6 1055 run without dflow

luitjansl:


rleander (approved):
Looks OK as far as I can see, I did not really check completeness.
To symplify the code I have the following recommendations:
* the decision whether or not to apply a mapping on a higher level:
Now the 'íf mapping defined' decision is inside the exchange function.
I would rather see that it is at the call side so that rge exchange function is NOT called when the mapping is not defined. The advantage is that the if..then are close together in the code and not scattered over routines, which makes it easier to oversee.
* instead of initializing mask and mapping in the dictionary to None and checking for None later, just DONT: a mapping is added to the dict when used and ommitted otherwise: On the check side, simply check of a certain label is in the dictionary, rather than checking for a None.
To be refactored later in a follow up issue.

PR #158, opened 2023-04-04
Imod6 1068 run tmodel f

HendrikKok:


imod_coupler/drivers/dfm_metamod/exchange.py
```diff
@@ -192,7 +192,7 @@ def set_realised_shortage_msw_mf6(
         shortage_left = shortage + demand_msw
         demand_mf6_negative = self.demand["mf-riv2dflow1d_flux_negative"]
         condition = np.logical_and(
```
rleander, line 194: condition is True when fully realized:
In that case sum_to_dflow>sum_from_dflow when both negative (extraction reported by dflowfm should exceed the extraction requested by the components). But always True when both are positive. Please rename variables to_dflow = requested, from_dflow is realized, also add this line as a comment. The condition is wrong when both are positive (false when sum_to_dflow<sum_from_dflow, but should be true) 
So you could also write:  sum_to_dflow>-abs(sum_from_dflow)

imod_coupler/drivers/dfm_metamod/dfm_wrapper.py
```diff
@@ -197,6 +215,20 @@ def get_1d_river_fluxes_ptr(self) -> NDArray[np.float_]:
...
+    def get_2d_river_fluxes_ptr(self) -> NDArray[np.float_]:
+        """
+        Returns
+        -------
+        Optional[NDArray[np.float_]]
+            the DFLOW_FM external fluxes ( "qext") for the 2d nodes
+        """
+        nr_nodes_1d = self.get_number_1d_nodes()
+        nr_nodes_2d = self.get_number_2d_nodes()
+        if nr_nodes_1d == 0:
```
rleander, line 227: should nr_nodes_2d be unequal to 0 and not nr_nodes_1d ?

rleander (changes requested):
Please have another look

luitjansl (commented):
you can make unit tests for the new wrapper functions. In a follow up task, if you want

rleander (approved):
Review OK

PR #186, opened 2023-08-22
Replace dfm update while loop

rleander:
replaced while loop for dfm timesteps by single call, can you plse review ?


HendrikKok (commented):
nice, but the tests fail

Hofer-Julian:
> nice, but the tests fail

The main pipeline doesn't work with lumbricus anymore.
You will have to fix https://dpcbuild.deltares.nl/buildConfiguration/iMOD6_Coupler_TestbenchCouplerWin64Lumbricus?mode=builds and use that instead.

PR #192, opened 2023-10-04
Disable `test_mf6_get_river_flux_estimate`

Hofer-Julian:
MODFLOW 6 internals changed, we should enable this test again as soon as possible


Hofer-Julian:
This doesn't cut it, the problem lies deeper.
Let's pin MODFLOW6 to last week's build for now.

PR #200, opened 2023-10-30
Lumbricus test2 d dfm2msw

rleander:


imod_coupler/drivers/dfm_metamod/dfm_metamod.py.orig
HendrikKok, line 1: this file could be removed i think

HendrikKok (commented):
This is the branch were we worked in together right? Should be all the work on making it work?

PR #201, opened 2023-10-30
Imod6 1082 modflow6 packages optional

rleander:
Please review: 
* packages in modflow now optional: that means if coupling not active, no need to have the package in the mf6 model
* pointers to kernel arrays now stored and retrieved from dictionaries
* active couplings listed in log


HendrikKok (changes requested):
Most of the changes seem ok. The weird thing is that all functions from line 479 of dfm_metamod.py seem to be changed (at least they are red) but they are not as far as I can see. Can you changes this maybe? For now its hard for me to review that part of the changes at this point.

PR #202, opened 2023-11-13
Create RibaMetaMod as a copy of RibaMod

HendrikKok:
Initial copy of RibaMod-class


HendrikKok:
Initial work on RibaMetaMod driver:

1 - Copy of Ribamod driver
2 - Copy of Ribamod tests
3 - Copy of Ribamod in primod

SonarCloud fails on code duplication: as expected since it’s a copy of Ribamod
Mypy fails on metamod: inherited this from main, not a result of work in this branch.

Hofer-Julian (approved):

PR #206, opened 2023-11-15
Update to newest Ribasim Python

Hofer-Julian:
That mostly involves the update to pydantic v2


Hofer-Julian:
@HendrikKok and @rleander, I've removed the ribametamod tests, since it turned out that some of the original ribamod tests were skipped and broke in the meantime. @Huite will have a look. After he fixed them, feel free to add corresponding ribametamod tests again.

PR #207, opened 2023-11-15
Fix ribamod tests

Hofer-Julian:
The glob in the ribamod tests was wrong, leading to all of them being skipped. In the meantime, they broke, so they need to be fixed now.


Hofer-Julian:
@rleander & @HendrikKok this should unblock ribametamod development

PR #215, opened 2023-11-27
Adapt to newest Ribasim version

Hofer-Julian:


HendrikKok (approved):
Nice work. I will adapt the temporary ribametamod tests later on

PR #219, opened 2023-12-14
Added tests for the excluding metaswap or ribasim in RibaMetaMod

rleander:
Added some testing for RibaMetaMod with meta or riba switched off, please review

### DIFF
--- a/imod_coupler/drivers/ribametamod/ribametamod.py
+++ b/imod_coupler/drivers/ribametamod/ribametamod.py
@@ -36,7 +36,7 @@ class RibaMetaMod(Driver):
     has_metaswap: bool  # configured with or without metaswap
 
     max_iter: NDArray[Any]  # max. nr outer iterations in MODFLOW kernel
-    delt: float  # time step from MODFLOW 6 (leading)
+    delt: float = 1.0  # time step from MODFLOW 6 (leading), default 1.0
 
     mf6_head: NDArray[Any]  # the hydraulic head array in the coupled model
     mf6_recharge: NDArray[Any]  # the coupled recharge array from the RCH package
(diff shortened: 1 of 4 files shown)


imod_coupler/drivers/ribametamod/ribametamod.py
```diff
@@ -36,7 +36,7 @@ class RibaMetaMod(Driver):
     has_metaswap: bool  # configured with or without metaswap
 
     max_iter: NDArray[Any]  # max. nr outer iterations in MODFLOW kernel
-    delt: float  # time step from MODFLOW 6 (leading)
+    delt: float = 1.0  # time step from MODFLOW 6 (leading), default 1.0
```
HendrikKok, line 39: this seems tricky. Normally MF updates this after the first init-timestep. Why this default than?

imod_coupler/drivers/ribametamod/ribametamod.py
```diff
@@ -259,6 +259,9 @@ def update(self) -> None:
             self.mf6.update()
 
         if self.has_ribasim:
+            # obtain modflow6 timestep, if not already done so
+            if self.delt is None:
```
HendrikKok, line 263: so the value is never updated by mf6 right? Since we use a default, it will never be None?

imod_coupler/drivers/ribametamod/ribametamod.py
```diff
@@ -267,13 +270,15 @@ def update(self) -> None:
...
         # exchange MODFLOW head to MetaSWAP
-        if self.has_metaswap:
-            self.exchange_mod2msw()
+        self.exchange_mod2msw()
 
-        # Do one MODFLOW 6 - MetaSWAP timestep
+        # obtain modflow6 timestep, if not already done so
         self.mf6.prepare_time_step(0.0)
-        self.delt = self.mf6.get_time_step()
+        if self.delt is None:
```
HendrikKok, line 279: again, how does this work with the default value?

HendrikKok (commented):
Most of it looks fine to me. I have some questions on the default value for delt.

PR #229, opened 2024-01-17
Add pixi task to publish primod

Hofer-Julian:


pixi.toml
```diff
@@ -32,12 +32,9 @@ format = "ruff format ."
...
 check-package-primod = { cmd = "rm --recursive --force dist && python -m build && twine check --strict dist/*", cwd = "pre-processing" }
-lint = { depends_on = [
-    "format",
-    "ruff",
-    "mypy-imodc",
-    "mypy-primod",
-] }
+lint = { depends_on = ["format", "ruff", "mypy-imodc", "mypy-primod"] }
+# Publish primod
+publish-primod = { cmd = "python -m build && twine check dist/* && twine upload dist/*", cwd = "pre-processing" }
```
Huite, line 37: Is there any chance at problems because there's an earlier existing dist directory?

E.g. if a module is deleted from one release to another, it wouldn't get overwritten if the new contents are copied to the dist dir.

Hofer-Julian: Good point, I now delete the dir first

Huite (approved):

PR #237, opened 2024-01-26
Testmodels two basin

Huite:
Moved the two_basin testmodel to ribasim-testmodels.


Hofer-Julian (approved):
After cleaning up the ruff lint, this is good to go from my side

PR #241, opened 2024-01-30
Improve debugging in visual studio code

### LINKED ISSUE #235: Document how to debug imod_coupler
When debugging imod_coupler, you basically have to replace `subprocess.run` with `__main__.run_coupler`.

We should document that process.

deltamarnix:
* Switch between `subprocess.run` and `imod_coupler.run_coupler()`, based on `pydevd` loaded or not.
* Add information in the README about debugging.
* Add some default settings for vs code that should be shared for pytest discovery.

Fixes #235

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,19 +5,19 @@ It currently focuses on groundwater and supports coupling between MetaSWAP and M
 
 It can be installed by running
 
-```
+```sh
 pip install imod_coupler
 ```
 
 Then you can run it as command line app via
 
-```
+```sh
 imodc /path/to/imod_coupler.toml
 ```
 
 In order to receive help for its usage, run
 
-```
+```sh
 imodc --help
 ```
 
@@ -33,7 +33,7 @@ In order to develop on `imod_coupler` locally, please follow the following steps
 
 - Create the environment by executing the following in your terminal:
 
-```
+```sh
 pixi run install
 ```
 
@@ -48,12 +48,13 @@ pixi run install
  - To run the tests it is advisable to have a `.env` file at the root of the project directory instead of modifying global environment variables. 
  The content of `.env` would then look similar to this with the variables `IMOD_COLLECTOR_DEVEL`, `IMOD_COLLECTOR_REGRESSION` and `METASWAP_LOOKUP_TABLE` adjusted to your local machine:
 
-```bash
+```sh
 IMOD_COLLECTOR_DEVEL='D:\checkouts\imod_collector_devel'
 IMOD_COLLECTOR_REGRESSION='D:\checkouts\imod_collector_regression'
 METASWAP_LOOKUP_TABLE='D:\checkouts\DSCtestbench\cases\e150_metaswap\f00_common\c00_common\LHM2016_v01vrz'
 
-IMOD_COUPLER_EXEC_DEVEL='imodc' # Specify an absolute path here to use a packaged version of iMOD Coupler
+# Specify an absolute path here to use a packaged version of iMOD Coupler
+IMOD_COUPLER_EXEC_DEVEL='imodc'
 IMOD_COUPLER_EXEC_REGRESSION='${IMOD_COLLECTOR_REGRESSION}/imod_coupler/imodc.exe'
 METASWAP_DLL_DEP_DIR_DEVEL='${IMOD_COLLECTOR_DEVEL}/metaswap'
 METASWAP_DLL_DEP_DIR_REGRESSION='${IMOD_COLLECTOR_REGRESSION}/metaswap'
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -69,12 +70,34 @@ RIBASIM_DLL_REGRESSION='${IMOD_COLLECTOR_REGRESSION}/ribasim/bin/libribasim.dll'
...
 ```
+
+- When developing with visual studio code, it is recommended to open the application via `open-vscode.bat`.
+  This will open the application in a new vscode window with the correct environment variables set.
+
+### Debugging
+
+When debugging the unit tests in visual studio code with the test explorer, you can encounter some problems.
+MODFLOW is not so keen on multi-threading, and when debugging a test, make sure that you only have one test running at a time.
+
```
Hofer-Julian, line 90:
```suggestion
Both MODFLOW 6 and MetaSWAP might behave unpredicateble when being initialized and finalized multiple times.
When debugging a test, make therefore sure that you only have one test running at a time.
```

README.md
```diff
@@ -69,12 +70,34 @@ RIBASIM_DLL_REGRESSION='${IMOD_COLLECTOR_REGRESSION}/ribasim/bin/libribasim.dll'
...
+When running the unit tests, this is not the case, since there is a switch statement that determines if we should call subprocess.run(), or stay within the main thread.
```
Hofer-Julian, line 91:
```suggestion
When you only run, not debug, the tests, this is not the case, since there is a switch statement that determines if we should call `subprocess.run()`, or stay within the main thread.
```

README.md
```diff
@@ -69,12 +70,34 @@ RIBASIM_DLL_REGRESSION='${IMOD_COLLECTOR_REGRESSION}/ribasim/bin/libribasim.dll'
...
+
+### Troubleshooting
+
+If you encounter errors while running the tests, it might be that your pip dependencies are outdated.
+This happens when you have pulled the latest changes from imod_coupler.
+In that case you need to update the pip dependencies as well.
+Try running:
+
+```sh
+pixi run install
```
Hofer-Julian, line 102:
```suggestion
pixi run update-git-dependencies
```

tests/fixtures/fixture_imodc_executable.py
```diff
@@ -0,0 +1,24 @@
...
+import pytest
+
+from imod_coupler.__main__ import run_coupler
+
+
+@pytest.fixture(scope="session")
+def imodc_executable(imod_coupler_exec_devel: Path) -> Callable[[Path], None]:
+    """Replacement for subprocess.run"""
+    if "pydevd" in sys.modules:
+
```
Hofer-Julian, line 15:
```suggestion
    if "pydevd" in sys.modules:
```

tests/fixtures/fixture_imodc_executable.py
```diff
@@ -0,0 +1,24 @@
...
+import pytest
+
+from imod_coupler.__main__ import run_coupler
+
+
+@pytest.fixture(scope="session")
+def imodc_executable(imod_coupler_exec_devel: Path) -> Callable[[Path], None]:
+    """Replacement for subprocess.run"""
+    if "pydevd" in sys.modules:
```
Hofer-Julian, line 14: What be good to add a comment, why we check for "pydevd" here

tests/test_parser.py
```diff
@@ -0,0 +1,12 @@
...
+import imod_coupler.parser
+from imod_coupler import __version__
+
+
+def test_get_version() -> None:
+    with pytest.raises(SystemExit) as cm:
+        output_version = imod_coupler.parser.parse_args(["--version"])
+        assert cm.value.code == 0
+        assert output_version is not None
+        assert output_version == __version__
```
Hofer-Julian, line 12: You expect an exception AND add assertions?
Are these assertions even checked then, or is the exception thrown when calling `parse_args`?

Hofer-Julian (commented):
Looks mostly good, I added a few comments.

tests/test_parser.py
deltamarnix: The thing is that the ArgumentParser, when performing the `version` action, actually also exits the whole program and that can be caught with a SystemExit. See: https://github.com/python/cpython/blob/c8cf5d7d148944f2850f25b02334400dd0238cb0/Lib/argparse.py#L2637C1-L2643C26

The test was not working, so I have updated the assertions to be outside of the `with` clause.

tests/fixtures/fixture_imodc_executable.py
deltamarnix: Auto-format doesn't like that

Hofer-Julian: Interesting. Well, then let's believe in auto-format :)

tests/fixtures/fixture_imodc_executable.py
```diff
@@ -0,0 +1,28 @@
...
+from collections.abc import Callable
+from pathlib import Path
+
+import pytest
+
+from imod_coupler.__main__ import run_coupler
+
+
+@pytest.fixture(scope="session")
+def imodc_executable(imod_coupler_exec_devel: Path) -> Callable[[Path], None]:
```
Hofer-Julian, line 12: I think this should be called `run_coupler` or something similar

deltamarnix: I went for run_coupler_function

Hofer-Julian (approved):
Debugging works as expected. Nice work @deltamarnix!
After the merge conflict is resolved, this is ready to go.

@rleander, @HendrikKok and @Huite this will be useful for you as well.

PR #243, opened 2024-02-01
Do not overwrite the drainage and infiltration values of uncoupled basins

Huite:
Pretty minimal, but maybe you reckon there's better naming possible.

This requires the changes of: https://github.com/Deltares/Ribasim/pull/1028

I will add a test based on the backwater model.

Will close:

https://github.com/Deltares/Ribasim/issues/998
https://github.com/Deltares/Ribasim/issues/999

### DIFF
--- a/imod_coupler/drivers/ribamod/ribamod.py
+++ b/imod_coupler/drivers/ribamod/ribamod.py
@@ -55,10 +55,14 @@ class RibaMod(Driver):
     ribasim_level: NDArray[Any]
     ribasim_infiltration: NDArray[Any]
     ribasim_drainage: NDArray[Any]
+    work_infiltration: NDArray[Any]
+    work_drainage: NDArray[Any]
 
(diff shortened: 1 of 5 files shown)


Huite:
I ended up adding tests to the two_basin model.
This currently tests for partial overlap, and no overlap at all. 

No overlap requires a check in primod for whether an exchange table has any entries or not.

I've minimized the stress periods for the two_basin cases to just a month rather than 10 years, which is obviously wasteful -- I picked to guarantee that it approaches a steady-state, but that's not really required for the plausibility tests here.

Changes here should also be reflected in the endtime of the `ribasim_testmodels.two_basin_model()`.

Huite:
I also wrote some utilities so that SonarCloud would stop bullying me about duplication

Hofer-Julian (approved):
Looks good to me!

PR #246, opened 2024-02-05
Set drainage and infiltration to NA for basins coupled to MODFLOW

Huite:
Fixes https://github.com/Deltares/Ribasim/issues/1073


Hofer-Julian (approved):

Huite:
This is currently failing on Ribasim allowing NaN/missing in basin.time but not basin.static.

PR #249, opened 2024-02-06
Update to pydantic 2.0 validators syntax

Hofer-Julian:
See https://docs.pydantic.dev/latest/migration/#changes-to-validators


Huite (approved):
Seems non-controversial!

PR #252, opened 2024-02-08
Move basin_definition from RibaMod to DriverCoupling

Huite:
This moves the basin_definition entry to the DriverCoupling. The idea here is to support multiple "stacked" basins: they share geometries, but not the basin node_ids.

This allows us to do the following:

```python
driver_coupling_1 = DriverCoupling(
    mf6_model="GWF_1",
    mf6_active_river_packages=["riv-1"],
    mf6_active_drainage_packages=["riv-1_drainage"],
    basin_definition=basin_definition_1,
)
driver_coupling_2 = DriverCoupling(
    mf6_model="GWF_1",
    mf6_active_river_packages=["riv-2"],
    mf6_active_drainage_packages=["riv-2_drainage"],
    basin_definition=basin_definition_2,
)

ribamod_model = RibaMod(
    ribasim_model=ribasim_model,
    mf6_simulation=mf6_simulation,
    coupling_list=[driver_coupling_1, driver_coupling_2],
)
```

This PR doesn't actually tests or something for this use of multiple definitions. The goal is to avoid API changes in the future, because we'd like this version to be used already with just a single basin_definition.
In principle, however, the code here should already work for multiple basin_definitions, connected to different MODFLOW 6 packages. In the future, we should probably think about some checks and add them. For example: the basin_definitions should not have overlapping node_id's between them.

This now also raises an error if there is no overlap found between Ribasim basins and MF6 boundaries. This allows us to get rid of several if-else's.

### DIFF
--- a/pre-processing/primod/ribamod/ribamod.py
+++ b/pre-processing/primod/ribamod/ribamod.py
@@ -27,6 +27,8 @@ class DriverCoupling:
     ----------
     mf6_model : str
         The model of the driver.
+    basin_definition : gpd.GeoDataFrame
+        GeoDataFrame of basin polygons
     mf6_active_river_packages : list of str
         A list of active river packages.
     mf6_passive_river_packages : list of str
@@ -38,6 +40,7 @@ class DriverCoupling:
     """
 
     mf6_model: str
+    basin_definition: gpd.GeoDataFrame
     mf6_active_river_packages: list[str] = field(default_factory=list)
     mf6_passive_river_packages: list[str] = field(default_factory=list)
     mf6_active_drainage_packages: list[str] = field(default_factory=list)
@@ -66,7 +69,6 @@ class RibaMod:
         ribasim_model: ribasim.Model,
         mf6_simulation: Modflow6Simulation,
         coupling_list: list[DriverCoupling],
-        basin_definition: gpd.GeoDataFrame,
     ):
         self.validate_time_window(
             ribasim_model=ribasim_model,
@@ -75,9 +77,6 @@ class RibaMod:
         self.ribasim_model = ribasim_model
         self.mf6_simulation = mf6_simulation
         self.coupling_list = coupling_list
-        if "node_id" not in basin_definition.columns:
-            raise ValueError('Basin definition must contain "node_id" column')
-        self.basin_definition = basin_definition
 
     def _get_gwf_modelnames(self) -> list[str]:
         """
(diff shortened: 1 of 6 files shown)


pre-processing/primod/ribamod/ribamod.py
```diff
@@ -89,6 +88,15 @@ def _get_gwf_modelnames(self) -> list[str]:
             if isinstance(value, GroundwaterFlowModel)
         ]
 
+    @staticmethod
+    def _new_coupling_dict() -> dict[str, Any]:
```
Hofer-Julian, line 92:
```suggestion
    def _empty_coupling_dict() -> dict[str, Any]:
```

pre-processing/primod/ribamod/ribamod.py
```diff
@@ -342,15 +357,55 @@ def write_exchanges(
...
-                    coupled_basin_indices.append(table["basin_index"])
 
-        if coupled_basin_indices:
-            coupled_basin_indices = np.unique(np.concatenate(coupled_basin_indices))
-            coupled_basin_node_ids = basin_ids[coupled_basin_indices]
-        else:
-            coupled_basin_node_ids = np.array([], dtype=int)
+                if table.empty:
+                    raise ValueError(
+                        f"No coupling can be derived for MODFLOW 6 package: {key}."
```
Hofer-Julian, line 363:
```suggestion
                        f"No coupling can be derived for MODFLOW 6 package: {key}. "
```

Hofer-Julian (commented):
Two small comments, otherwise looks good.

Huite:
What's debatable about this PR: this is arguably abusing the DriverCoupling concept (?).
In this case, there are multiple driver couplings that are merged together, because they belong to a single MF6 GWF model.
As far as I can tell, the DriverCouper idea is for multi-model support. In this case, keeping them separate will now require changes to the driver, as none of the drivers will currently accept multiple driver couplings.

However, this is in this case, by far the most pragmatic solution. Alternatives are:

* Adding multiple columns with MF6 package names to the basin definition. This is quite an ugly solution.
* Some additional level of indirection: more nested, more grouped, more mappings. They all end up being more confusing.

Finally, the nested for loop is what is it is. The data here are necessarily nested, so I think it's okay for the code to reflect that.

pre-processing/primod/ribamod/ribamod.py
Huite: Good call

HendrikKok (approved):
Looks fine by me. Is there an issue already on the future tests for this type of coupling?

Huite:
No issues yet. Feel free to open some (and assign yourself ^^)!

PR #256, opened 2024-02-13
Update numerical scheme riba meta mod

HendrikKok:
fixes https://github.com/Deltares/Ribasim/issues/971


imod_coupler/drivers/ribametamod/exchange.py
```diff
@@ -0,0 +1,200 @@
...
+        This function computes the realised (negative) volumes
+        """
+        shortage = np.absolute(self.demand - realised)
+        demand_negative = self.demand_negative
+        self._check_valid_shortage(shortage, demand_negative)
+        # deal with zero division
+        realised_fraction = np.where(
+            demand_negative < 0.0, 1.0 - (-shortage / demand_negative), 1.0
+        )
+        for flux_label in self.flux_labels:
```
rleander, line 38: It can be shown that we can savely do:
realised_fraction = np.where(
            self.demand_negative < 0.0, 
            (realised - self.demand + self.demand_negative) / self.demand.negative,
            1.0)
and by defiinition self.demand is less or equal realised
and self.demand equal realised if self.demand_negative is zero

Huite (approved):

rleander (approved):
We can only evaluate this numerical scheme in the debugger with a testmodel (which is in main) ,  even more when the currently missing parts of the ribasim exchange are complete. We should approve now, merge it and then create a new branch from main

PR #259, opened 2024-02-19
Additionally check if columns are in data frame

HendrikKok:
fixes: 258

### DIFF
--- a/pre-processing/primod/ribamod/ribamod.py
+++ b/pre-processing/primod/ribamod/ribamod.py
@@ -283,18 +283,23 @@ class RibaMod:
 
         Ribasim will otherwise overwrite the values set by the coupler.
         """
+
         # FUTURE: in coupling to MetaSWAP, the runoff should be set nodata as well.
+        def _nullify(df: pd.DataFrame) -> None:
+            """Set drainage and infiltration columns to nodata if present in df"""
+            if df is not None:
+                columns_present = list(
+                    {"drainage", "infiltration"}.intersection(df.columns)
+                )
+                if len(columns_present) > 0:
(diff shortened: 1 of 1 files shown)


HendrikKok:
All tests are green, but I don’t know if it is correct that the static data frame is optionally None. In that case we lose some asserts in the tests.

tests/test_primod/test_preprocessing_ribamod.py
```diff
@@ -250,14 +250,19 @@ def test_nullify_on_write(
...
+            df.loc[df["node_id"] == 2, ["drainage", "infiltration"]]
+            .isna()
+            .all(axis=None)
+        )
+        # Basin 3 is uncoupled, terms should be preserved.
+        assert (
+            df.loc[df["node_id"] == 3, ["drainage", "infiltration"]]
+            .notna()
+            .all(axis=None)
+        )
```
Hofer-Julian, line 265: I think this part should be removed.
Agreed @Huite?

Hofer-Julian (changes requested):

tests/test_primod/test_preprocessing_ribamod.py
Huite: I would suggest setting a value for the dataframe instead in the test.

I think it would be fine to test it twice: once with the model as is (which shouldn't raise errors) and a second time, after which you set the dataframe with e.g. zero values for all columns. Then do these asserts.

HendrikKok: Done

Hofer-Julian (approved):

PR #261, opened 2024-02-26
Move from requests to httpx

Hofer-Julian:
It is better maintained

Also see this blog post by one of the maintainers of requests: https://blog.ian.stapletoncordas.co/2024/02/a-retrospective-on-requests


deltamarnix (approved):
I'm ok with moving to this package, as it's only a scripts package. Consider if you want to move it to `build-dependencies`, since it's not a dependency that someone would need in order to use imod_coupler.

I know this is something new and we don't use it yet, so leaving it like this is alright with me.

Issue #262, opened 2024-02-26
primod.RechargeSvatMapping doesn't drop layer dimension


JoerivanEngelen:
The RechargeSvatMapping requires a 2D field of active Recharge cells which need to be coupled to SVATs.
The recharge grid can be provided in two ways:

1. A 2D grid with dims (y, x)
2. A 3D grid with dims (layer, y, x), where the layer dimension is length one.

In the latter case, we call xarray ``.drop_vars("layer")``  to drop the layer dimension.
However, I'm getting reports this doesn't actually drop the dimension, but creates an empty dimension, which still needs to be squeezed.
Furthermore I see we have not separate unit tests for this mapping to test its validation.

Huite:
I think the right method here would've been `.squeeze("layer", drop=True)` instead.

PR #263, opened 2024-02-26
Issue #262 primod rchsvatmapping layer dim

JoerivanEngelen:
This PR fixes issue #262, as well as adds some unit tests to verify that most common cases in which recharge can be provided work.

- 2D grid (y, x) with ``layer`` coord (succeeds)
- 3D grid with layer dimension, length 1 (succeeds)
- 3D grid with multiple layers (fails)
- 2D transient grid (fails)

I placed the new tests in the ``tests/test_mapping.py``, but I noticed there is a ``imod_coupler/mapping`` namespace, as well as a ``primod/mapping`` namespace. I was a bit in doubt whether mixing these in ``tests/test_mapping.py`` is a good idea. If you have ideas for a better location to put these tests, let me know.

### DIFF
--- a/pre-processing/primod/mapping/rch_svat_mapping.py
+++ b/pre-processing/primod/mapping/rch_svat_mapping.py
@@ -39,12 +39,13 @@ class RechargeSvatMapping(MetaModMapping):
         super().__init__()
         self.dataset["svat"] = svat
         self.dataset["layer"] = xr.full_like(svat, 1)
-        if "layer" in recharge.dataset.coords:
-            self.dataset["rch_active"] = (
-                recharge.dataset["rate"].drop_vars("layer").notnull()
-            )
+        rate = recharge.dataset["rate"]
+        if "layer" in rate.dims:
+            rate_no_layer = rate.squeeze("layer", drop=True)
         else:
-            self.dataset["rch_active"] = recharge.dataset["rate"].notnull()
+            # 'layer' can be still in a coord, so ensure is also dropped
+            rate_no_layer = rate.drop_vars("layer", errors="ignore")
+        self.dataset["rch_active"] = rate_no_layer.notnull()
         self._pkgcheck()
         self._create_rch_id()
 
--- a/tests/fixtures/fixture_modflow.py
+++ b/tests/fixtures/fixture_modflow.py
@@ -149,13 +149,16 @@ def inactive_idomain() -> xr.DataArray:
 
     return idomain
 
+
 @pytest_cases.fixture(scope="function")
 def recharge(active_idomain: xr.DataArray) -> mf6.Recharge:
     return make_recharge(active_idomain)
 
+
 @pytest_cases.fixture(scope="function")
 def coupled_mf6_model(active_idomain: xr.DataArray) -> mf6.Modflow6Simulation:
     return make_coupled_mf6_model(active_idomain)
+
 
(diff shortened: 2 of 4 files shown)


JoerivanEngelen:
I see some tests are labeled as "flaky", making it appear as if this is not my wrongdoing. Is this the case or is it likely I introduced errors in RibaMod?

Hofer-Julian (approved):
I am not super familiar with that part of the codebase, but the changes look good to me.

The testbench was red, but I think that was just because of a version mismatch between Ribasim and Ribasim Python. If it is green now, we can merge this from my side.

pre-processing/primod/mapping/rch_svat_mapping.py
```diff
@@ -39,12 +39,13 @@ def __init__(self, svat: xr.DataArray, recharge: mf6.Recharge):
         super().__init__()
         self.dataset["svat"] = svat
         self.dataset["layer"] = xr.full_like(svat, 1)
-        if "layer" in recharge.dataset.coords:
-            self.dataset["rch_active"] = (
-                recharge.dataset["rate"].drop_vars("layer").notnull()
-            )
+        rate = recharge.dataset["rate"]
+        if "layer" in rate.dims:
+            rate_no_layer = rate.squeeze("layer", drop=True)
```
luitjansl, line 44: should there be a check that the length of the layer dimension is 1?

tests/fixtures/fixture_modflow.py
```diff
@@ -65,6 +65,15 @@ def make_mf6_simulation(gwf_model: mf6.GroundwaterFlowModel) -> mf6.Modflow6Simu
     return simulation
 
 
+def make_recharge(idomain: xr.DataArray) -> mf6.Recharge:
```
luitjansl, line 68: make_recharge_package?

tests/fixtures/fixture_modflow.py
```diff
@@ -65,6 +65,15 @@ def make_mf6_simulation(gwf_model: mf6.GroundwaterFlowModel) -> mf6.Modflow6Simu
...
+    idomain_l1 = idomain.sel(layer=1)
+    recharge = xr.zeros_like(idomain_l1, dtype=float)
+    recharge[:, 0] = np.nan
```
luitjansl, line 71: are you assigning nan to all x for y =0? (if so, why is that? if something else, what exactly?)

luitjansl (approved):
approved (but some comments/questions)

pre-processing/primod/mapping/rch_svat_mapping.py
JoerivanEngelen: The squeeze method automatically throws an error if length is more than 1.

tests/fixtures/fixture_modflow.py
JoerivanEngelen: The opposite, all cells in the first column (x) are assigned ``np.nan``. This is consistent with the MetaSWAP model, where these cells are inactive, and thus no mapping should take place to these cells.

JoerivanEngelen: I added a comment to clarify

tests/fixtures/fixture_modflow.py
JoerivanEngelen: Renamed

PR #267, opened 2024-03-11
changes in ribamod for changed names in Ribasim API

rleander:
name change in ribamod because of change in ribasim api


HendrikKok:
names are fine, but try to fix the failing tests as well.

Huite (approved):

PR #268, opened 2024-03-11
Pin pixi for now

Hofer-Julian:
Let's revert this when https://github.com/prefix-dev/pixi/pull/951 is merged and released


rleander (approved):
splendid! thx!

PR #273, opened 2024-03-19
Refactor drivercoupling

Huite:


pre-processing/primod/driver_coupling/metamod.py
```diff
@@ -0,0 +1,117 @@
...
+from imod.mf6 import GroundwaterFlowModel, Modflow6Simulation
+from imod.msw import GridData
+
+from primod.driver_coupling.driver_coupling_base import DriverCoupling
+from primod.mapping.node_svat_mapping import NodeSvatMapping
+from primod.mapping.rch_svat_mapping import RechargeSvatMapping
+from primod.mapping.wel_svat_mapping import WellSvatMapping
+
+
+@dataclass
```
Hofer-Julian, line 14: I think we should go for pydantic instead

Huite: I concur

pre-processing/primod/coupled_model.py
```diff
@@ -0,0 +1,53 @@
+import abc
+from collections.abc import Sequence
+from pathlib import Path
+from typing import Any
+
+from primod.driver_coupling.driver_coupling_base import DriverCoupling
+
+
+class CoupledModel(abc.ABC):
+    coupling_list: Sequence[DriverCoupling]
```
Hofer-Julian, line 10: mypy probably made you do this, but `Sequence` for an actual variable feels weird to me.
I would use `Sequence` only for function parameters and actual types like `list` for variables, but if `mypy` is the reason you did this you can leave it as is.

pre-processing/primod/driver_coupling/metamod.py
```diff
@@ -0,0 +1,106 @@
...
+            )
+        elif not sprinkling_in_msw and sprinkling_in_mf6:
+            raise ValueError(
+                f"Modflow 6 Well package {self.mf6_wel_package} specified for sprinkling, "
+                "but no Sprinkling package found in MetaSWAP model."
+            )
+        elif sprinkling_in_msw and sprinkling_in_mf6:
+            return True
+        else:
+            return False
```
Hofer-Julian, line 52: That use case is the reason why `match` exists nowadays :)

pre-processing/primod/driver_coupling/ribameta.py
```diff
@@ -0,0 +1,148 @@
...
+    Attributes
+    ----------
+    basin_definition: gpd.GeoDataFrame
+        GeoDataFrame of basin polygons
+    user_demand_definition: gpd.GeoDataFrame
+        GeoDataFrame of user demand polygons
+    """
+
+    ribasim_basin_definition: gpd.GeoDataFrame
+    ribasim_user_demand_definition: gpd.GeoDataFrame | None
```
Hofer-Julian, line 32:
```suggestion
    ribasim_user_demand_definition: gpd.GeoDataFrame | None = None
```

You have to set a default value if you want to make it optional. Otherwise, you have to specify `None` when constructing it. If that's what you want you can ignore me

pre-processing/primod/mapping/svat_user_demand_mapping.py
```diff
@@ -0,0 +1,37 @@
...
+        svat: xr.DataArray,
+        index: NDArray[Int],
+    ):
+        condition = svat.notnull()
+        user_id = xr.where(condition, gridded_user_demand, np.nan)  # type: ignore
+        include = user_id.notnull().to_numpy()
+        user_id_values = user_id.to_numpy()[include].astype(int)
+        user_index = np.searchsorted(user_demand_ids, user_id_values)
+
+        # TODO: I'm not entirely sure this is the correct logic!
```
Hofer-Julian, line 25:
```suggestion
        # TODO (Huite): I'm not entirely sure this is the correct logic!
```

Hofer-Julian (approved):
This cleans things up quite a bit. Great job @Huite!

I've added a few minor comments, after they are resolved this is good to go from my side.

pre-processing/primod/coupled_model.py
Huite: Yeah it's due to mypy, since it complains that lists are invariant.
To be fair, any sequence also actually works here given the logic, which is just a loop over the driver couplings.

pre-processing/primod/driver_coupling/ribameta.py
Huite: Good catch!

PR #289, opened 2024-04-30
Fix tests mismatch riba mod coordinates

rleander:
We need this to make the ribamod tests run again

### DIFF
--- a/tests/test_imod_coupler/test_ribamod.py
+++ b/tests/test_imod_coupler/test_ribamod.py
@@ -148,23 +148,21 @@ def test_ribamod_backwater(
     riv0 = results.budgets["riv-1"].isel(time=-2).compute()
 
     # Get the last flow between the edges
-    final_flow = results.flow_df[results.flow_df["time"] == "2020-12-31"]
+    volume_balance = results.basin_df[results.basin_df["time"] == "2020-12-31"]
     # Check's what lost and gained in the basins
     basin_ids = ribamod_model.ribasim_model.basin.node.df["node_id"]
     ribasim_budget = (
-        final_flow.loc[
-            final_flow["from_node_id"].isin(basin_ids)
-            & final_flow["to_node_id"].isin(basin_ids)
-        ]["flow_rate"]
-        * 86_400.0
+        (volume_balance['inflow_rate'] - 
+         volume_balance['outflow_rate'] - 
+         volume_balance['balance_error'])[:-1] * 86_400.0
     ).to_numpy()
     # It seems that there is a match of integrated volumes shifted by half a timestep.
     # the integration of flux on the Ribasim side deserves a closer look
     modflow_budget = (
         (0.5 * drn + 0.5 * riv + 0.5 * drn0 + 0.5 * riv0).sel(y=0).to_numpy()
     )
-    budget_diff = ribasim_budget + modflow_budget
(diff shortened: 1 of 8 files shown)


tests/test_imod_coupler/test_ribamod.py
```diff
@@ -135,7 +135,7 @@ def test_ribamod_backwater(
         run_coupler_function=run_coupler_function,
     )
 
-    final_level = results.basin_df[results.basin_df["time"] == "2021-01-01"]["level"]
+    final_level = results.basin_df[results.basin_df["time"] == "2020-12-31"]["level"]
```
HendrikKok, line 138: Why do you change this? You extend the model period, but take the water level after one year?

tests/test_imod_coupler/test_ribamod.py
```diff
@@ -148,23 +148,23 @@ def test_ribamod_backwater(
     riv0 = results.budgets["riv-1"].isel(time=-2).compute()
 
     # Get the last flow between the edges
-    final_flow = results.flow_df[results.flow_df["time"] == "2020-12-31"]
+    volume_balance = results.basin_df[results.basin_df["time"] == "2020-12-31"]
```
HendrikKok, line 151: again?

tests/test_imod_coupler/test_ribamod.py
```diff
@@ -148,23 +148,23 @@ def test_ribamod_backwater(
...
         * 86_400.0
     ).to_numpy()
     # It seems that there is a match of integrated volumes shifted by half a timestep.
     # the integration of flux on the Ribasim side deserves a closer look
     modflow_budget = (
         (0.5 * drn + 0.5 * riv + 0.5 * drn0 + 0.5 * riv0).sel(y=0).to_numpy()
     )
-    budget_diff = ribasim_budget + modflow_budget
-    assert (np.abs(budget_diff) < 1.0e-6).all()
+    budget_diff = ribasim_budget - modflow_budget
```
HendrikKok, line 166: The Ribasim budget now has a different sign?

HendrikKok (commented):
I have some questions on the period extension you did and the evaluation of the results. These seem inconsistent to me now.

tests/test_imod_coupler/test_ribamod.py
rleander: There is a problem with the first day of the new year, which seems just beyond the range of acceptable entries, that is why the final_level was inquired a day earlier, just at the end of the simulation. The extension with two years was only for the bucket model (because the bucket seems to empty slower now) and not for the backwater model.

tests/test_imod_coupler/test_ribamod.py
rleander: This is the corresponding volume balance, the other was the level

tests/test_imod_coupler/test_ribamod.py
rleander: The ribasim budget now determined in a totally different way. was previously determined by comparing flows on links and now on the change in node volumes. 
modflow budgets are outflow for modflow,  the ribasim budget is the ribasim inflow, taking the balance error into account. Both budgets should ideally be equal now

HendrikKok (approved):

PR #296, opened 2024-05-24
Ribasim1164 mf6 correction

rleander:
replace the dummy values in the realised infiltration by the real values obtained from the integrated drainage and infiltration from Ribasim

### DIFF
--- a/imod_coupler/drivers/ribametamod/ribametamod.py
+++ b/imod_coupler/drivers/ribametamod/ribametamod.py
@@ -195,6 +195,12 @@ class RibaMetaMod(Driver):
             # Get all Ribasim pointers, relevant for coupling with MODFLOW 6
             self.ribasim_infiltration = self.ribasim.get_value_ptr("basin.infiltration")
             self.ribasim_drainage = self.ribasim.get_value_ptr("basin.drainage")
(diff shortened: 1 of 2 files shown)


imod_coupler/drivers/ribametamod/ribametamod.py
```diff
@@ -382,15 +388,17 @@ def update(self) -> None:
...
-        if self.has_ribasim and self.has_metaswap:
-            self.update_ribasim_metaswap()
-        elif self.has_ribasim and not self.has_metaswap:
-            self.update_ribasim()
-
         if self.has_ribasim:
-            # get realised values on basin boundary nodes and exchange correction flux
-            realised_basin_nodes = self.exchange.demand  # dummy value for now
-            self.exchange.to_modflow(realised_basin_nodes)
+            self.ribasim_infiltration_sum[:] = 0.0
```
HendrikKok, line 392: Can you move this to an exchange method?

rleander: done, moved it to the exchange.demand method, so that immediately after setting the demands to ribasim, its integrated volumes are zeroed

HendrikKok (approved):

PR #297, opened 2024-05-28
Ribasim1417_check_only_drn_in_passive

rleander:


pre-processing/primod/driver_coupling/ribamod.py
```diff
@@ -106,12 +106,22 @@ def write_exchanges(
                     save_flows=True,
                 )
                 destination = "river"
+                if isinstance(self, RibaModPassiveDriverCoupling):
+                    raise TypeError(
+                        f"Expected Drainage (no 'active' packages in passive coupling), received: {type(package).__name__}"
```
HendrikKok, line 111: "Expected Drainage packages for passive coupling,  received: {type(package).__name__}" ?

tests/test_primod/test_preprocessing_ribamod.py
```diff
@@ -141,7 +143,7 @@ def test_ribamod_write_toml(
     mf6_modelname, mf6_model = get_mf6_gwf_modelnames(mf6_bucket_model)[0]
     mf6_river_packages = get_mf6_river_packagenames(mf6_model)
 
-    driver_coupling = RibaModPassiveDriverCoupling(
```
HendrikKok, line 144: Maybe it’s better to change the package, now you’re changing the complete test? Is there still passive coupling being tested here?

rleander: I am not sure I understand this comment. 
The bucket model does not seem to have drainage, only a river package, which raises an exception now.

rleander: Changed the river package in the model to a drain and changed the active drivercoupling back to a passive drivercoupling as it was. Also removed the corresponding api-package associated with the river.
Explicit assert on the exception and the message introduced in this issue.

HendrikKok (approved):

PR #300, opened 2024-06-05
fixed typo in var name

rleander:
realized and realised mixed up here


rleander:
minor stuff, but just not so nice ... and confusing

HendrikKok (approved):

PR #304, opened 2024-06-20
check on riv and multiple subgrid

rleander:


HendrikKok (approved):
Looks fine by me. We could change one line if you agree.

PR #307, opened 2024-06-28
maximize realised array

HendrikKok:


imod_coupler/drivers/ribametamod/ribametamod.py
```diff
@@ -472,6 +472,13 @@ def exchange_sprinkling_flux_realised_msw2rib(self, delt: float) -> None:
 
         nonzero_user_indices = np.flatnonzero(self.mapped_sprinkling_demand)
 
+        # maximize realised array by demand values
+        too_large = np.flatnonzero(
+            (self.ribasim_user_realized / days_to_seconds(self.delt_sw))
+            > self.mapped_sprinkling_demand
+        )
+        self.ribasim_user_realized[too_large] = self.mapped_sprinkling_demand[too_large]
+
```
rleander, line 481: * self.mapped_sprinkling_demand is een flux, volume per time, whereas ribasim_user_realized is a volume.
So the comparison too_many = [...] is dimensionally correct, but the assignment not, I think.
should it not be: 
self.ribasim_user_realized[too_large] = self.mapped_sprinkling_demand[too_large] *  days_to_seconds(self.delt_sw)
* maybe shorter:
self.ribasim_user_realized[:] = np.minimum(self.ribasim_user_realized[:],
	self.mapped_sprinkling_demand * days_to_seconds(self.delt_sw)) 
and skip the 'too_large' array

rleander (changes requested):

imod_coupler/drivers/ribametamod/ribametamod.py
HendrikKok: Your right indeed. Now using np.minimum()

rleander (approved):

PR #314, opened 2024-07-12
Update RibaMetaMod-tests with cases where dtsw != dtgw

HendrikKok:


rleander (approved):
we need the fix in the main urgently, do deeper analysis later

PR #317, opened 2024-08-06
Find library in path

rleander:
No longer needed to fully specify the path to the kernel libraries (the leaf 'dll' in the toml files), when it is referred to from the PATH environment variable, hence allowing the use of environment modules on linux. Library locations are resolved upon readin the config, where applicable, whereas explicit paths are left as they are


imod_coupler/drivers/driver.py
```diff
@@ -3,11 +3,26 @@
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any
-
+import os, sys
 from loguru import logger
 
 from imod_coupler.config import BaseConfig
 
+def resolve(libname):
```
Huite: `match ... case` works quite nicely here:

```python
import os
import sys
from pathlib import Path

def resolve(libname):
    if os.path.isfile(os.path.abspath(libname)):
        return os.path.abspath(libname)

    match sys.platform.lower():
        case "win32":
            env_var = 'PATH'
        case "linux" | "linux2":
            env_var = 'LD_LIBRARY_PATH'
        case "darwin":
            env_var = 'LD_LIBRARY_PATH'
        case _:
            return libname  # Unknown platform, return original libname

    pathdef = os.environ.get(env_var, '')
    
    for dir in pathdef.split(os.pathsep):
        full_path = Path(dir) / libname
        if os.path.isfile(full_path):
            return str(full_path)
    
    return libname  # if resolution failed, give it back to the call site
```

imod_coupler/drivers/driver.py
```diff
@@ -9,6 +11,26 @@
...
+        case _:
+            return libname
+
+    if os.path.isfile(libname):
+        return libname
+    if env_var in os.environ:
+        pathdef: str = os.environ[env_var]
+        for dir in pathdef.split(os.pathsep):
+            full_path = Path(dir) / libname
+            if os.path.isfile(full_path):
```
Huite, line 29: You can just use `full_path.is_file()` here right? Since `is_file()` is a method on Path objects.

Huite (approved):

Huite:
Minor nitpick still -- I approved it anyway.

PR #328, opened 2024-09-12
fix issue and add test

HendrikKok:


tests/test_imod_coupler/test_ribametamod_cases.py
```diff
@@ -326,6 +328,59 @@ def case_bucket_model(
     )
 
 
+def case_bucket_model_no_subgrid(
```
rleander, line 331: Is the case of basins coupled to metaswap but not to modflow also tested here  ?

rleander (changes requested):

tests/fixtures/fixture_ribasim.py
```diff
@@ -36,6 +36,13 @@ def ribasim_bucket_model() -> ribasim.Model:
     return add_subgrid(bucket)
 
 
+@pytest_cases.fixture(scope="function")
+def ribasim_bucket_model_no_subrgid() -> ribasim.Model:
```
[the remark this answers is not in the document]
HendrikKok: done

tests/test_imod_coupler/test_ribametamod_cases.py
HendrikKok: No that is indeed not tested for now.

rleander (approved):
Prima werk!

PR #329, opened 2024-09-12
removed redundant ribamod write call

rleander:
test fails on write protection trying to write a ribasim toml file that is still open.
removed superfluous write call ...


visr:
> test fails on write protection trying to write a ribasim toml file that is still open

That should be fixed with https://github.com/Deltares/Ribasim/pull/1831

PR #331, opened 2024-09-30
Pass metaswap timestep to update_ribasim

visr:
This seems to match the signature better.

https://github.com/Deltares/imod_coupler/blob/162db85368122e9e03c319a96a51ee9bf1d75789/imod_coupler/drivers/ribametamod/exchange.py#L186


visr:
This was correct since this is only called if there is no metaswap. Changed it into a comment instead.

HendrikKok (approved):

PR #335, opened 2024-10-07
Test ribasim integration take 2

visr:
This is like #334 but I changed or removed the last 5 commits there, to remove the timeouts and target https://github.com/Deltares/Ribasim/pull/1874 instead.

You may still want to modify or rename `ribasim_user_realized` since that now indeed is cumulative. On the Ribasim I renamed the variable to `user_demand.cumulative_inflow` for clarity. I guess `ribasim_user_realized_save` is no longer needed?


visr:
My modifications have been incorporated into #334 which has been merged, so closing.

PR #336, opened 2024-10-08
Allow grid based basin and water user definitions and facilitate subsetting of subgrid-df

Huite:
The driver_couplings related to Ribasim can now contain grid-based basin and water user definition. This gives the optional ```xr.Dataset``` or ```xr.DataArray```.

- ```xr.Dataset``` in ```primod.driver_coupling.RibaModDriverCoupling```:  per mf-package  a basin definition data variable
```xr.DataArray```  in ```primod.driver_coupling.RibaMetaDriverCoupling```: per (single) basin definition and (single) water user definition.

In the ```primod.driver_coupling.ActiveNodeBasinMapping``` an optional subsetting is possible to deal with stacked subgrid points. Subsetting is done via the optional ‘meta_label’ column in the subgrid.df

### DIFF
--- a/pre-processing/primod/driver_coupling/ribameta.py
+++ b/pre-processing/primod/driver_coupling/ribameta.py
@@ -15,6 +15,7 @@ from primod.driver_coupling.util import (
 )
 from primod.mapping.svat_basin_mapping import SvatBasinMapping
 from primod.mapping.svat_user_demand_mapping import SvatUserDemandMapping
+import xarray as xr
 
 
 class RibaMetaDriverCoupling(DriverCoupling):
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: pre-processing/primod/driver_coupling/ribameta.py [hunk 1]
```diff
@@ -28,8 +29,8 @@ class RibaMetaDriverCoupling(DriverCoupling):
         GeoDataFrame of user demand polygons
     """
 
-    ribasim_basin_definition: gpd.GeoDataFrame
-    ribasim_user_demand_definition: gpd.GeoDataFrame | None = None
+    ribasim_basin_definition: gpd.GeoDataFrame | xr.DataArray
```
Huite, line 32: Should be `gpd.GeoDataFrame | xr.Dataset` right, not DataArray?
With the variable names matching the MF6 keys.

HendrikKok: For MetaSWAP coupling we need a single basin definition. A ``` xr.DataArray ``` would be sufficient right?

rleander (approved):

Issue #342, opened 2024-10-25
primod: Update mapping objects to update to iMOD Python's refactoring plans


JoerivanEngelen:
I need to refactor iMOD Python's MetaSWAP objects, see https://github.com/Deltares/imod-python/issues/1255. This has the following consequences for ``primod``:

Part 1 of this refactoring consists of changing the well argument for the coupling objects from ``imod.mf6.wel.WellDisStructured`` to ``imod.mf6.mf6_wel_adapter.Mf6Wel``. The latter is a direct data representation of the MODFLOW 6 WEL file. More can be read here https://github.com/Deltares/imod-python/issues/728. This has the following consequences for iMOD Coupler:

- The ``imod.mf6.mf6_wel_adapter.Mf6Wel`` object needs to be provided to coupling objects as argument instead of the deprecated ``imod.mf6.wel.WellDisStructured``.
- ``WellSvatMapping.self._create_well_id`` needs to be updated
- Test fixtures need to be updated.

Part 2 is a larger refactoring, where this ``Mf6Wel`` object will be an argument of the ``Sprinling.write`` and ``CouplerMapping.write`` methods, instead of the initialization. More can be read here https://github.com/Deltares/imod-python/issues/1255. This will require the following updates to ``primod``:

- In ``MetaMod.write``, Mf6Wells need to be fetched from ``self.mf6_simulation`` and provided to ``self.metaswap.write``.
- In ``MetaModDriverCoupling.derive_coupling``, Mf6Wells need to be fetched from ``gwf_model``
- Test fixtures need to be updated.
- More?

As I need to run ``primod`` against a dev version of iMOD Python, this also means I need to temporarily modify the pixi.toml, to use a install from git with pip, instead of a install from conda-forge.

Issue #333, opened 2024-10-02
Primod's basin mapping can't deal with integer basin id's with latest Rasterio dependecy 


HendrikKok:
problem: imod.prepare.rasterize results in a crash if node_id is an integer and fill-value is the default np.nan. 

solution: We could use a fill-value of 0 if node_id is integer but this does not work in combination with NodeBasinMapping logic. Best is to force node_id to float64 dtype.

JoerivanEngelen:
FYI: This is an issue with rasterio >=1.3.10.

Setting fillid to -1 caused the tests to succeed, but I don't know if this aspect is well tested.

JoerivanEngelen:
FYI, as it is related: In this PR https://github.com/Deltares/imod_coupler/pull/348 I had to force the dtype argument for imod.prepare.rasterize to float, otherwise problems arose. 

I recently updated iMOD Python to take a ``dtype`` argument for its ``rasterize`` function, but haven't released this yet.

PR #348, opened 2024-10-28
Issue #344 part1 imod refactor primod update

### LINKED ISSUE #344: Update primod to keep up with iMOD Python's MetaSWAP refactoring part 1
Part 1 of this refactoring consists of changing the well argument for the coupling objects from ``imod.mf6.wel.WellDisStructured`` to ``imod.mf6.mf6_wel_adapter.Mf6Wel``. The latter is a direct data representation of the MODFLOW 6 WEL file. More can be read here https://github.com/Deltares/imod-python/issues/728. This has the following consequences for iMOD Coupler:

- The ``imod.mf6.mf6_wel_adapter.Mf6Wel`` object needs to be provided to coupling objects as argument instead of the deprecated ``imod.mf6.wel.WellDisStructured``.
- ``WellSvatMapping.self._create_well_id`` needs to be updated
- Test fixtures need to be updated.

JoerivanEngelen:
UPDATE: This is a new attempt for a PR, in the other one I accidentily merged locally and pushed to feature branch. I reset that now.

Closes #344, this is the first task to update ``primod`` for the iMOD Python refactor.

Plans are layed out in #342. In this first task, I changed the use of the deprecated ``WellDisStructured`` object, to the Mf6Wel object, which is a direct translation of MODFLOW 6's cellids to a file. As this iMOD Python refactor is still work in progress, I temporarily introduce a install-imod task, which pip installs imod from a git branch, instead of a release on conda-forge. After I have accomplished part 2 of the refactors, I'll make an iMOD Python release, and this can be reverted back.

It does the following things:

- Temporarily introduce a install-imod task, which install imod from a git branch
- Use the ``Mf6Wel`` object instead of the ``WellDisStructured`` object
- I had to drop an unused layer coordinate in a RibaMetaMod fixture
- I force the ``dtype`` in calls to imod.prepare.rasterize to float, to work around this issue https://github.com/Deltares/imod_coupler/issues/333. The ``dtype`` argument is only included in the the dev version of iMOD Python yet.


pixi.toml
```diff
@@ -29,6 +28,11 @@ scipy = "*"
 tomli = "*"
 tomli-w = "*"
 xmipy = "*"
+filelock = "*"
```
Manangka, line 31: Could you sort the dependencies?

Manangka (approved):

pre-processing/primod/mapping/wel_svat_mapping.py
```diff
@@ -71,7 +73,7 @@ def _create_well_id(
         layer = np.tile(well_layer, (n_subunit, 1))
         layer_1d = layer[well_active]
 
-        well_id = self.well.dataset.coords["index"] + 1
+        well_id = well_cellid.coords["ncellid"] + 1
```
HendrikKok, line 76: ncellid is actually an index right? otherwise the +1 is now unnecessary.

tests/fixtures/fixture_ribametamod.py
```diff
@@ -17,7 +17,7 @@ def make_msw_model(
 
     idomain = gwf["GWF_1"]["dis"]["idomain"]
     if "layer" in idomain.dims:
-        idomain_layer1 = idomain.sel(layer=1)
+        idomain_layer1 = idomain.sel(layer=1, drop=True)
```
HendrikKok, line 20: why is this?

HendrikKok (commented):
Nice work, I had a few comments.

pixi.toml
JoerivanEngelen: I defer from that now, as this is a temporary measure. After iMOD Python has been released with the API update, I will revert these changes to the pixi.toml. Instead, I will then pin iMOD Python to be above or higher a specific version number.

pre-processing/primod/mapping/wel_svat_mapping.py
JoerivanEngelen: That is correct!

tests/fixtures/fixture_ribametamod.py
JoerivanEngelen: I ran into an issue with some tests, where iMOD Python was complaining about faulty coordinates. It turned out as the layer coord was passed on to ``imod.util.spatial.get_cell_area``, which throws an error if there is any coord more than x, y, dx, and dy. This could be easily fixed by dropping the layer coordinate upon selecting the first layer in the fixture. It did not seem like the layer coord was necessary in the first place, so I think dropping it is a slight improvement regardless.

Issue #345, opened 2024-10-25
Update primod to keep up with iMOD Python's MetaSWAP refactoring part 2


JoerivanEngelen:
Part 2 is a larger refactoring, where this ``Mf6Wel`` object will be an argument of the ``Sprinling.write`` and ``CouplerMapping.write`` methods, instead of the initialization. More can be read here https://github.com/Deltares/imod-python/issues/1255. This will require the following updates to ``primod``:

- In ``MetaMod.write``, Mf6Wells need to be fetched from ``self.mf6_simulation`` and provided to ``self.metaswap.write``.
- In ``MetaModDriverCoupling.derive_coupling``, Mf6Wells need to be fetched from ``gwf_model``
- Test fixtures need to be updated.
- More?

PR #351, opened 2024-11-07
Issue #345 part2 imod refactor primod update

JoerivanEngelen:
Fixes https://github.com/Deltares/imod_coupler/issues/345

This PR introduces the following things to primod:

- Provide MODFLOW6 objects in primod at write, instead of at init.
- Introduce a ``ModflowMixin``, which can be used to add methods for MODFLOW-related couplings. At the moment all couplings use MODFLOW, but there might be a Ribasim-WFLOW coupling in the future, for which MODFLOW methods are unnecessary. It now contains one static method.
- Add unittests for ModflowMixin
- Update fixtures and tests
- Update a fixture which calls an internal key "max_abstraction_surfacewater_m3_d", which is changed to "max_abstraction_surfacewater" in iMOD Python

- Boyscouting: Change calls from _get_pkg_key("dis") to _get_dis_key(), to also support getting imod.mf6.VerticesDiscretization in the future, for unstructured grids.


JoerivanEngelen:
I see ribasim python crashing with Python 3.10 on a Pandera issue, I think this is irrelevant to this changeset. This also happened with the previous changeset. I'll fix this in the final merge.

pre-processing/primod/model_mixin.py
```diff
@@ -0,0 +1,27 @@
...
+
+from imod.mf6 import Modflow6Simulation, StructuredDiscretization
+from imod.mf6.mf6_wel_adapter import Mf6Wel
+
+
+class ModflowMixin:
+    """Modflow specific methods to add to coupling objects."""
+
+    @staticmethod
+    def get_mf6_pkgs_with_coupling_dict(
```
HendrikKok, line 11: Could you make this name more specific? We also couple to RIV, DRN and API packages in the Ribasim coupling for example. Those are not included here. Perhaps something like: get_mf6_pkgs_for_metaswap_coupling

tests/test_imod_coupler/test_ribametamod_cases.py
```diff
@@ -88,12 +87,8 @@ def set_confined_storage_formulation(
     return mf6_model
 
 
-def remove_sprinkling_from_groundwater(
-    msw_model: MetaSwapModel, mf6_model: Modflow6Simulation
-) -> MetaSwapModel:
+def remove_sprinkling_from_groundwater(msw_model: MetaSwapModel) -> MetaSwapModel:
     msw_model.pop("sprinkling")
-    msw_model.pop("mod2svat")
-    msw_model["mod2svat"] = msw.CouplerMapping(modflow_dis=mf6_model["GWF_1"]["dis"])
```
HendrikKok, line 96: Why did you remove this? Since the CouplerMapping was initialized with sprinkling activated, it includes those couplings. Thats why it is reinitialized.

HendrikKok (changes requested):
Nice work, I have one small request on a name, and one question. Also, the test bench is not green right?

tests/test_imod_coupler/test_ribametamod_cases.py
JoerivanEngelen: Because the CouplerMapping doesn't take any arguments anymore. The modflow packages are now requested upon writing, instead of upon initialization, in iMOD Python. In pseudo-code:

Before the refactor in iMOD Python:

```python
class CouplerMapping:
    def __init__(self, modflow_dis, modflow_wel=None):
         ...
    def write(self):
         ...    
```

After it:

```python
class CouplerMapping:
    def __init__(self):
         ...
    def write(self,modflow_dis, modflow_wel=None):
         ...    
```

Therefore the CouplerMapping object doesn't need to be re-initialized: mappings are created upon writing.

pre-processing/primod/model_mixin.py
JoerivanEngelen: Good point! I didn't consider the Ribasim coupling when naming the method. Will fix.

JoerivanEngelen:
> Nice work, I have one small request on a name, and one question. Also, the test bench is not green right?

See my previous comment ;-)

> I see ribasim python crashing with Python 3.10 on a Pandera issue, I think this is irrelevant to this changeset. This also happened with the previous changeset. I'll fix this in the final merge of the feature branch to main.

pre-processing/primod/model_mixin.py
JoerivanEngelen: I followed your suggestion for the method name, and renamed ModflowMixin to MetaModMixin, as the method is not used for RibaMod.

HendrikKok (approved):

PR #353, opened 2024-11-08
Limit imod python version to <0.18.0, before refactor is merged

JoerivanEngelen:
The 0.18.0 release of iMOD Python will feature a significant refactoring of the MetaSWAP objects in iMOD Python. 
Once 0.18.0 is released, the fixes here https://github.com/Deltares/imod_coupler/pull/352 will update primod and the test fixtures to be functional again.

The 0.18.0 release will break the current main, so to ensure it does not when iMOD Python 0.18.0 is released. I limit it in the ``pixi.toml`` here to keep the CI stable. This gives me the time to get to work https://github.com/Deltares/imod_coupler/pull/352 .


JoerivanEngelen:
I discussed with @HendrikKok, the release test job is failing, but this is due to changes outside this PR. In the release test job, and old version of Ribasim is used, which isn't able to compute water balances without errors larger than 1e-4.

PR #406, opened 2026-01-13
Add user acceptance test

JoerivanEngelen:
This PR adds a user acceptance test, namely the LHM. The data is stored in a minio S3 bucket and version managed with dvc. Similar to the approach taken in iMOD Python. The only data excluded from the data version control is the 80 GB MetaSWAP database, but this is included on the MinIO bucket, unzipped. Before running the user acceptance tests, users have to mount the bucket, so that the path to the MetaSWAP database is provided to the S3 bucket. 

The README has instructions on how to run the user acceptance tests locally. Note: running the user acceptance tests locally requires an ethernet connection for optimal performance. Running the test on WiFi in Utrecht took 1.5 hours, whereas running it with an ethernet connection in Delft took 45 minutes.

The user_acceptance tests are pytest marked, and their fixtures are set to ``autouse=False``, so that the tests will only be run when explicitly asked for, and the heavy fixtures will only be generated when asked for (i.e. when running the user acceptance tests).

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -9,6 +9,7 @@ environments:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/noarch/_python_abi3_support-1.0-hd8ed1ab_2.conda
       - conda: https://conda.anaconda.org/conda-forge/noarch/adwaita-icon-theme-49.0-unix_0.conda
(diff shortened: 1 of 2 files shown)


tests/test_user_acceptance/test_metamod_lhm.py
```diff
@@ -0,0 +1,559 @@
...
+        "rchindex2svat.dxc",
+        "wellindex2svat.dxc",
+    }
+    assert set(dxc_filenames) == expected_dxc_filenames
+
+    path_nodenr2svat = written_lhm_conversion / "exchanges" / "nodenr2svat.dxc"
+    path_rchindex2svat = written_lhm_conversion / "exchanges" / "rchindex2svat.dxc"
+    path_wellindex2svat = written_lhm_conversion / "exchanges" / "wellindex2svat.dxc"
+
+    settings = {"delimter": "\s+", "index_col": False}
```
rleander73: plse fix typo, this one on two other lines in this file ..

rleander73 (approved):
all ok!

Issue #465, opened 2026-04-30
Release iMOD Collector zip to Github releases


Manangka:
Add a deploy step that creates a collector release on Github and hook it up to the Deploy pipeline

Manangka:
Completed

PR #467, opened 2026-04-30
Issue #465 Add the github release pipeline

Manangka:
In this PR a github release pipeline is added to the deploy pipeline.

It creates a new release on Github and adds the sources and collector.zip to it.
The imodc.exe in the released is signed.
It extracts the release notes from the changelog.md file and adds it to the release description

The readme.me had been extended with instructions on how to create a release

### DIFF
--- a/.teamcity/Deploy/DeployProject.kt
+++ b/.teamcity/Deploy/DeployProject.kt
@@ -1,7 +1,7 @@
 package Deploy
 
+import IMODCollector.buildTypes.IMODCollector_X64development
 import _Self.vcsRoots.ImodCoupler
-import jetbrains.buildServer.configs.kotlin.AbsoluteId
 import jetbrains.buildServer.configs.kotlin.BuildType
 import jetbrains.buildServer.configs.kotlin.FailureAction
 import jetbrains.buildServer.configs.kotlin.Project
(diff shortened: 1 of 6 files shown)


.teamcity/Deploy/DeployProject.kt
```diff
@@ -1,15 +1,78 @@
...
+        }
+    }
+
+    dependencies {
+        dependency(AbsoluteId("SigningAndCertificates_IMOD_SigningCollector")) {
+            snapshot {
+                onDependencyFailure = FailureAction.FAIL_TO_START
+            }
+
+            artifacts {
```
HendrikKok, line 68: This can potentially go wrong since it does not check if the collector belongs to the specific release right? Or is this covered in the singing build?

Manangka: It is covered by TeamCity.
The way Teamcity works is that the commit/tag used to the toplevel job is passed along to the jobs upstream of it.
In this case the signing job will be triggered with the same commit sha which in turns calls the collector job with the same commit sha

HendrikKok: ok, clear

HendrikKok (approved):

PR #479, opened 2026-05-19
Add continue logic for non-convergence 

### LINKED ISSUE #476: Coupler will always continue in case of non-convergence
The expected behavior would be:
1- Continue in cases the `continue` option is activated in MODFLOW
2- Stop in all other cases.

The current iteration scheme relies on the BMI to crash in case of non-convergence. So in the past the coupler would always stop.
In the current situation the BMI does not crash anymore and the coupler will always continue.

 We should use the `continue` variable of the `ims` input of MODFLOW to evaluate if we need to continue in case of non-covnvergence. 

The `solve` should look like:

```python

    def solve(self,) -> None:
        self.mf6.prepare_solve(1)
        for _ in range(1, self.max_iter + 1):
            has_converged = self.mf6.solve(1)
            if has_converged:
                break
        if not has_converged and not self.mf6.continue_solve:
            raise Exception("Solution did not converged; simulation aborted")
        self.mf6.finalize_solve(1)



```

HendrikKok:
should fix https://github.com/Deltares/imod_coupler/issues/476

#### Summary

This PR adds support for the MODFLOW 6 `CONTINUE` option in the iMOD Coupler, allowing coupled simulations to proceed even when the solver does not converge. 

#### `imod_coupler`
- The coupler now checks the `CONTINUE` flag from the MODFLOW 6 simulation.
- If `CONTINUE` is `true`, a warning is logged when convergence is not reached and the simulation proceeds.
- If `CONTINUE` is `false` (default), a `RuntimeError` is raised on non-convergence, ans simulation is aborted

#### Tests (`test_metamod.py`)
- Added `test_metamod_solve_failure`: verifies that a deliberately non-converging simulation (tight solver settings) raises `CalledProcessError`, and that the same simulation completes successfully when `CONTINUE` is set in `mfsim.nam`.
- Added helper `add_continue_to_mfsim_nam`: reads and patches `mfsim.nam` to insert `CONTINUE` into the `OPTIONS` block. This workaround is needed because imod-python does not yet support writing the `CONTINUE` option.

#### Notes
- The manual file patching in the test (`add_continue_to_mfsim_nam`) should be replaced once imod-python exposes this option in its API.


tests/test_imod_coupler/test_metamod.py
```diff
@@ -410,6 +410,67 @@ def test_metamod_exchange_logging(
...
+    """
+    content = mfsim_path.read_text()
+    lines = content.splitlines(keepends=True)
+
+    # Check if OPTIONS block already exists
+    begin_idx = next(
+        (i for i, line in enumerate(lines) if line.strip().upper() == "BEGIN OPTIONS"),
+        None,
+    )
+    if begin_idx is not None:
```
Manangka: This if else statement is not really needed is it? When is the Option block not present?

If you get rid of the logic this method simplifies to 

```
    content = mfsim_path.read_text()
    lines = content.splitlines(keepends=True)

    begin_idx = [line.strip().upper() for line in lines].index("BEGIN OPTIONS")
    end_idx = [line.strip().upper() for line in lines].index("END OPTIONS", begin_idx)
    block = [line.strip().upper() for line in lines[begin_idx + 1 : end_idx]]
    
    if "CONTINUE" not in block:
            lines.insert(begin_idx + 1, "  CONTINUE\n")
    
    mfsim_path.write_text("".join(lines))
```

Manangka (changes requested):

HendrikKok:
@copilot resolve the merge conflicts in this pull request

Manangka (approved):

Manangka (approved):

PR #480, opened 2026-05-20
Fix silent crash when the run_coupler encountered an exception

Manangka:
The exception raised when a dll crashed was silently ignored when calling it though the run_coupler_debug method.

This PR fixes that. The ProcessPoolExecutor automatically handles exception unlike multiprocessing.Process.
The exception is then rethrown as a CalledProcessError exception to be in line with the exception that would be raised when it was run through the imodc.exe


rleander73 (approved):
nice work!

PR #483, opened 2026-06-04
Update to Ribasim v2026.1.1

### LINKED ISSUE #431: Test Ribasim v2026.1.0-rc2 with iMOD-coupler
I understood that you are only testing tagged releases. Since we intend this to be the last pre-release before a stable release, could you make sure it works? I think that there is at least one thing to update; `init_julia` no longer needs to be called explicitly, see https://github.com/Deltares/Ribasim/commit/1fe6468d29ca2edec14e39d7fef1a67ba2aa69be#diff-539e5efe040987eeb0eab970222595d2ce0de87bfa17f1353c0b3b8a87f13d10

Here is the release:
https://github.com/Deltares/Ribasim/releases/tag/v2026.1.0-rc2

Not sure if that affects you, but note that `ribasim.exe` moved into the `bin` dir, next to `libribasim.dll`.

Happy to help. It should be done in March still so we have time to fix possible regressions in Ribasim.

### LINKED ISSUE #484: Remove pytest-dotenv plugin
It is unmaintained and has been archived a year ago: https://github.com/quiqua/pytest-dotenv
We can probably ensure the .env file is loaded in our tests ourselves.

visr:
See https://github.com/Deltares/Ribasim/releases/tag/v2026.1.1

This release makes allocation use adaptive timestepping by default, just like the physical layer.
That means users run into infeasible optimization problems less.

https://github.com/Deltares/Ribasim/releases/tag/v2026.1.1

Fixes https://github.com/Deltares/imod_coupler/issues/431
Fixes #484 in 122b0c2d736e36f859d363c95dd1dd6298c8074b by calling `load_dotenv(override=True)` ourselves.
Closes #482 since this also updates the lock file. We had to implement pandas 3 support via small changes like 509d53f2e004a7550602a6088839b951600b61c5.


visr:
@rleander73 this is ready for review. The two allocation test failures are now marked as xfail. I confirmed these models fail (or even hang) on the v2026.1.1 release, but work fine on main.

HendrikKok (approved):

PR #492, opened 2026-07-14
Parallel coupler for metamod (revised)

verkaik:


imod_coupler/drivers/ribametamod/ribametamod.py
verkaik, line 1: Is this a bug in the main?!

PR #495, opened 2026-08-04
Ensure that the Pixi pipeline fails when one of its substeps fail

Manangka:
The pixi pipeline failed however on TeamCity it is still marked as succeeded.
This is because the powershell version in which the pipeline is run doesn't handle errors correctly.

In this PR the powershell version being used is changed to  7.3 Core. This version has some build in error handling settings that we can use. In particular:
- PSNativeCommandUseErrorActionPreference: This is used to capture the error code of the commands being called in the script e.g. "pixi run -e pixi-update update"
- ErrorActionPreference: This determines the action after an error occured

After these changes the build should fail "correctly"


Manangka:
Converted to draft. The correct PowerShell version isn't available on the agents

Edit:
The required powershell version has been made available in #498 498

PR #514, opened 2026-09-16
Issue #513 Create wellindex that is contiguous

### LINKED ISSUE #513: primod writes incorrect wellindex2svat.dxc to fail for test case with inactive MODFLOW6 cells
primod can create arrays with non-contiguous well indexes that go out of range for cases with an inactive idomain. This causes a fail when trying to run the iMOD Coupler

JoerivanEngelen:
Fixes #513

This fixes the test case where non-contigous well indices were written by primod to wellindex2svat.dxc. This causes a crash when trying to run iMOD Coupler.

Before this fix the well index written to wellindex2svat in the "inactive" metamod_case was:

``array([ 1,  2,  3,  4,  5,  6,  7,  9, 10, 11, 12, 13, 14, 15])``

Where the new one is now:

``array([ 1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14])``


tests/test_primod/test_preprocessing_metamod.py
```diff
@@ -172,6 +172,38 @@ def test_metamod_write_exchange_no_sprinkling(
...
+    exchange_dir = output_dir / "exchanges"
+    wel_dxc = fixed_format_parser(
+        exchange_dir / WellSvatMapping._file_name,
+        WellSvatMapping._metadata_dict,
+    )
+
+    n_mf6_wells = coupled_mf6_model_inactive["GWF_1"]["wells_msw"].dataset.sizes[
+        "index"
+    ]
+    assert max(wel_dxc["wel_id"]) <= n_mf6_wells
```
rleander73: would expect something like:
assert_arrays_equal(
    np.sort(np.array(wel_dxc["wel_id"]),
    np.arange(n_mf6_wells)+1)

JoerivanEngelen: Right, I'll adapt the assertion statement, it indeed matches the PR description better

pre-processing/primod/mapping/wel_svat_mapping.py
```diff
@@ -73,7 +73,11 @@ def _create_well_id(
         layer = np.tile(well_layer, (n_subunit, 1))
         layer_1d = layer[well_active]
 
-        well_id = well_cellid.coords["ncellid"] + 1
```
HendrikKok, line 76: Thats a bug indeed. We need the internal index of the well-package array, not the one of the global MF6-grid

HendrikKok (approved):

PR #516, opened 2026-09-17
Add sprinkling points test case

JoerivanEngelen:
Fixes this issue on the iMOD Python board: https://github.com/Deltares/imod-python/issues/1908

This PR:

- Adds test case which uses the new iMOD Python SprinklingPoints feature. In this case all svats (of both subunits) are connected to a single well.
- Modifies the assertion of fluxes somewhat so that it both works for metaswap models built with the SprinklingGrid as well as SprinklingPoints functionality.


JoerivanEngelen:
> ### 🔵 Needs a closer look
> ### Suppressed comments (2)
> **tests/fixtures/fixture_metaswap.py:283**
> 
> * `active_idomain` is a MODFLOW mask, so after expanding it this assigns sprinkling point 1 to every cell in both subunits, including the cells that are not MetaSWAP SVATs (`GridData.active` is false at x=100 and `area` is NaN for one subunit in two rows). `SprinklingPoints` treats those extra art-grid links as surface-water connections, so this case does not represent only the stated groundwater links to all actual SVATs and can introduce unintended surface-water demand. Build `art_grid` from the MetaSWAP grid's area/active mask instead.
> 
> ```
>     art_grid = active_idomain.sel(layer=1, drop=True).expand_dims(subunit=[0, 1])
> ```
> 
> **tests/test_imod_coupler/test_metamod.py:218**
> 
> * This comparison reduces the point case to a spatial total, so it cannot detect an incorrect point-to-well mapping: with only one well, the test would still pass if the flux were written to a different cell or layer. Please add an assertion that the nonzero MODFLOW 6 flux is at the well's `(x_p, y_p, layer_p)` location (and zero elsewhere), while retaining the total comparison for the point case.
> 
> ```
>         np.testing.assert_allclose(
>             msw_sprinkling_fluxes_m3.sum(dim=["y", "x"]).data,
>             mf6_sprinking_fluxes.sum(dim=["y", "x"]).data,
> ```
> 

I've dealt with these comments: The spatial distribution of sprinkling points is now tested as well. And made sure ``art_grid`` received the right masking.

ClaireDons (approved):
Looks good to me (although I'm not familiar with the codebase so perhaps missed something).

PR #519, opened 2026-09-18
update artifacts config (#518)

HendrikKok:


.teamcity/_Self/buildTypes/Lint.kt
rleander73, line 26: Why store environments in this directory? Is it always accessible on TC agents? Are re-using predefined environments this way?

